Issue #12, opened 2018-07-23
Make Rule of Merging Pull Request


ynamiki:
We need to make a rule of merging a pull request.
I think a simple rule is sufficient for this time.
@sergefdrv, @Naoya-Horiguchi, please give me your comments and suggestions.

- A pull request must be reviewed by one of the nec-blockchain members (@sergefdrv, @Naoya-Horiguchi and @ynamiki).
- The reviewer merges the request if it is fine.
- If the contributor (i.e. a creator of a Pull Request) is a member of nec-blockchain, the request must be reviewed and is merged by another member.

sergefdrv:
I am absolutely fine with these rules :+1:

nhoriguchi:
+1

ynamiki:
Thanks, let's start with the rules.
If you are inconvenienced in practice, please create an issue to update the rule.

sergefdrv:
I'm wondering how to handle updates to pull request based on review comments. I'm not really happy that GitHub in contrast to GitLab doesn't allow to see different versions of rebased and force-pushed commit series :slightly_frowning_face: It looks like the expected way to address review comments is to add additional commits on top. I believe this will make history cluttered and less useful. I was used to a more old-fashion workflow:

- Send a patch series to the mailing list, with a cover letter explaining the whole series. This is analogous to submitting a pull request in GutHub
- Collect review comments and discuss.
- Overwrite the commits that need to be changed. Drop commits that do not make sense. Create new commits only if that makes the series to a more compelling story.
- Send revised commit sequence as another patch series to the mailing list, labelling it with version number, like v2, v3 etc. And repeat the review-revise-resend as necessary.
- When the series is good enough to merge, it gets applied to the master branch, with no merge commit.

What I liked in that workflow is that it produced a clean readable history of changes applied to the mater branch. This history could be referred to later to easily find the meaningful commit that made a certain change, e.g. using `git blame`. I also liked to see a patch/commit series as a compelling story telling how the proposed bigger change is achieved with smaller, conceivable steps.

I'm afraid that addressing review comments in separate commits would introduce too much noise in the history and make it less usable. On the other hand, rewriting commits with `git rebase` and `git push --force` on a MR in GutHub seem to hide reviewer comments on the older versions of the commits and the commits might be shown in a weird order. 

[...]

PR #14, opened 2018-07-26
Remove USIG epoch handling from core code

sergefdrv:
The replicas need to agree so that each peer has a single instance of USIG able to produce valid UIs. This is required to prevent a faulty replica from sending conflicting messages to different nodes in the consensus network. The current approach of capturing the USIG epoch value from the first valid UI received from a peer relies on a number of assumptions:
- all replicas are initially correct,
- each replica uses its unique USIG key pair per consensus protocol instance,
- the first UI is generated using a single USIG instance per replica,
- those first UIs are received and processed by all correct replicas before any replica becomes faulty in a sense that it starts generating and sending UIs using another USIG instance initialized with the same sealed key pair.

Those assumptions might be too strong in some environments. In that case, there should be a way for all correct replicas to agree on a single USIG instance identity per each replica and use that identity to verify UIs of received consensus messages.

Considering this, it seems to be most reasonable to move the current approach of achieving the agreement on USIG identity outside of the core protocol implementation. 

This series does some refactoring to make it easier and eventually moves the USIG epoch handling from the core protocol implementation to the sample implementation of the external authentication interface.


sergefdrv:
@luthlee Could you please review this PR?

sergefdrv:
@ynamiki Somehow I don't seem to be able to request review from luthlee...

image

sergefdrv:
@ynamiki I wonder why this PR doesn't go through CI?

sergefdrv:
@Naoya-Horiguchi @ynamiki Who of you would also like to review this PR?

nhoriguchi (dismissed):

nhoriguchi:
uh, I pushed review button in order to approve the first commit of this series, but a whole pull request was approved which is not yet intended. I don't see how to cancel my approval, but anyway I still continue to review to this.

ynamiki:
> Somehow I don't seem to be able to request review from luthlee...

@sergefdrv [Reviewers need to have read access to the repository](https://help.github.com/articles/requesting-a-pull-request-review/). We have two options for give the access: 1) add to organization `nec-blockchain` (if luthlee is a member of your lab), or 2) add as [an outside collaborator](https://help.github.com/articles/adding-outside-collaborators-to-repositories-in-your-organization/) of this repository `nec-blockchain/minbft`. Which is preferred?

ynamiki:
> I wonder why this PR doesn't go through CI?

@sergefdrv I have update settings in CircleCI to build forked pull requests. It will work at next commit (I have tested with #15).

nhoriguchi:
@sergefdrv, I don't get into details (I need learn existing code more..), I have a few feedbacks:

- patch descriptions are nicely written and I felt easy to understand what you try to do.
- you state in comment around usig_create_ui() "the epoch and counter values in little-endian byte order", I guess this is because it's required by hardware. If so, that's worth explaining in the comment.
- time.Sleep() inserted for simulation mode had better be conditional by checking environment variable SGX_MODE.

usig/sgx/sgx-usig_test.go
```diff
@@ -53,6 +53,9 @@ func TestSGXUSIG(t *testing.T) {
 	assert.NotNil(t, usig, "Got nil SGXUSIG instance")
 	defer usig.Destroy()
 
+	usigID, err := MakeID(usig.Epoch(), usigPubKey)
+	require.NoError(t, err)
+
```
luthlee: Unit test for the `ParseID()`

sergefdrv:
The commits are shown in the pull request by GitHub in a weird way because of https://help.github.com/articles/why-are-my-commits-in-the-wrong-order/, https://github.com/isaacs/github/issues/386.

sample/authentication/crypto.go
```diff
@@ -122,19 +125,38 @@ func (a *PublicAuthenScheme) VerifyAuthenticationTag(m []byte, sig []byte, pubKe
 	return nil
 }
 
-// SGXUSIGAuthenticationScheme impelements AuthenticationScheme
-// interface by utilizing SGX USIG to create/verify authentication
-// tags.
+// usigKeyFingerprint is a SHA256 hash of the USIG public key.
+type usigKeyFingerprint [sha256.Size]byte
```
luthlee: Maybe you want to use a shorter hash output or truncate the output? SGX ECDSA is 256-bit, so the public key is 512-bit long. Your fingerprint only saves half the space.

sergefdrv:
I rebased the commits with `--ignore-date` and they appear in the right order on GitHub. https://github.com/isaacs/github/issues/386#issuecomment-319142149

sergefdrv:
> We have two options for give the access: 1) add to organization nec-blockchain (if luthlee is a member of your lab), or 2) add as an outside collaborator of this repository nec-blockchain/minbft. Which is preferred?

@ynamiki Please go ahead with option (1). She is from our lab :slightly_smiling_face:

sergefdrv:
>     you state in comment around usig_create_ui() "the epoch and counter values in little-endian byte order", I guess this is because it's required by hardware. If so, that's worth explaining in the comment.

That is not really the requirement. It was just easier to do so. SGX is only available on Intel CPUs. Those are always little-endian. The signature payload is conveniently constructed as a "packed" C structure (see https://github.com/sergefdrv/minbft/blob/bf79a36638d99873ae5c28141c8410c7e4771abe/usig/sgx/enclave/usig.c#L30). That is why those numbers are in little endian. Changing to big endian would require adding more code in enclave, which is better to keep at minimun.

>     time.Sleep() inserted for simulation mode had better be conditional by checking environment variable SGX_MODE.

`SGX_MODE` determines the enclave build mode, but only at build time. It is not reliable indicator at runtime. For instance, `bin/keytool generate` mentioned in https://github.com/nec-blockchain/minbft#generating-keys can be invoked in a separate shell with no `SGX_MODE` set.

I was thinking of building both version of enclave and store them in separate files. Then one could chose the version to use at runtime. What do you think?

sergefdrv:
@Naoya-Horiguchi Please do not merge this PR until @luthlee has finished reviewing it.

sample/authentication/crypto.go
```diff
@@ -163,10 +185,45 @@ func (au *SGXUSIGAuthenticationScheme) VerifyAuthenticationTag(m []byte, sig []b
...
+	// using a single USIG instance per replica. Moreover, those
+	// first UIs are assumed to be received and processed by
+	// correct replicas before any replica becomes faulty in a
+	// sense that it starts generating and sending UIs using
+	// another USIG instance initialized with the same sealed key
+	// pair.
+	//
+	// Those assumptions might be too strong in some environments.
+	// In that case, there should be a way for all correct
+	// replicas to agree on a single USIG instance identity per
```
luthlee: If there is a correct bootstrap/initialization stage, replicas are not required to be all correct. 
Therefore, "in that case, a correct bootstrap is required for all replicas to first agree on...".

sergefdrv:
> time.Sleep() inserted for simulation mode

What we could also do is to make a pull request to change the way SGX simulation mode sets the seed value. If it gets accepted, we could get rid of this hack when the fix gets released. https://github.com/intel/linux-sgx/issues/246#issuecomment-408167720

sergefdrv:
> I don't see how to cancel my approval

@Naoya-Horiguchi You could try this https://blog.github.com/2016-10-12-dismissing-reviews-on-pull-requests/

ynamiki:
> Please go ahead with option (1). She is from our lab 🙂

I have invited @luthlee to `nec-blockchain`.

usig/sgx/sgx-usig_test.go
sergefdrv: Thanks to pointing that out. I'll add a test for this.

sample/authentication/crypto.go
sergefdrv: I'm thinking to truncate this to 64-bit fingerprint.

sample/authentication/crypto.go
sergefdrv: Exactly, there would be no need for all replicas to be initially correct. That is why only the correct replicas need to agree; up to f replicas could be faulty from the beginning and do not agree. I didn't want to emphasize how it should be achieved. It could be achieved at bootstrap phase or maybe some other way. But I'm considering to change the wording to: "In that case, all correct replicas are required to agree ..." Would that make sense?

sergefdrv:
@luthlee I have tried to address your comments. Sorry, I didn't know how GitHub handles it and force-pushed the changed series here. So there seem to be no way to compare with the previous revision. Next time I will follow some other approach as discussed in #12 .

luthlee (approved):

sergefdrv:
@Naoya-Horiguchi Would you have more comments?

nhoriguchi:
> I was thinking of building both version of enclave and store them in separate files. Then one could chose the version to use at runtime. What do you think?

I think that if this is easy enough, you can do it.

> What we could also do is to make a pull request to change the way SGX simulation mode sets the seed value. If it gets accepted, we could get rid of this hack when the fix gets released. intel/linux-sgx#246 (comment)

That's the best scenario for us, so I hope your suggestion will be accepted.

sergefdrv:
@Naoya-Horiguchi Building both HW and simulation version of enclave would require significant change in enclave Makefile. This looks like a separate issue to address later. 

As of changing SGX SDK code, it doesn't look like they are going to change it soon. We could propose the change ourselves. This is rather easy change, but somebody would need to take care of this.

sergefdrv:
@Naoya-Horiguchi Actually, changing the way the enclave is build wouldn't eliminate this workaround with 1 second delay. We would still use simulation mode in CI, for example. So the best way would be to propose the change to SGX SDK.

sergefdrv:
@Naoya-Horiguchi 

> because updated version will be posted

Sorry, the content of this pull request has been updated since your comment https://github.com/nec-blockchain/minbft/pull/14#issuecomment-408305157.  Next time I will follow the approach as discussed in #12 .

nhoriguchi:
@sergefdrv thank you for the explanation. It seems to me that the RDRAND approach pointed out in the SGX SDK thread might be fine because code change is minimum (maybe calling sgx_read_rand() to generate a seed?) and we might avoid additional library dependency as in nanosecond approach. RDRAND is available since Ivy Bridge CPU, so most of our platform should have it.

sergefdrv:
@Naoya-Horiguchi 
> It seems to me that the RDRAND approach pointed out in the SGX SDK thread might be fine 

I'm not sure if RDRAND would always be available in a could-based CI. 

> because code change is minimum (maybe calling sgx_read_rand() to generate a seed?)

We do not generate the seed ourselves, this is what SGX SDK does. We use `sgx_read_rand()` https://github.com/nec-blockchain/minbft/blob/master/usig/sgx/enclave/usig.c#L171 and `sgx_ecc256_create_key_pair()` https://github.com/nec-blockchain/minbft/blob/master/usig/sgx/enclave/usig.c#L119. The point is that any random number SGX SDK provides in simulation mode is based on a seed value captured at enclave creation time. It is currently the local time in seconds. That is SGX SDK implementation.

> and we might avoid additional library dependency as in nanosecond approach. 

I'm not sure what do you mean by additional library dependency. I was suggesting if some of us could prepare a patch for SGX SDK to use higher-precision timestamp for seeding the pseudo random number generation in simulation mode.

> RDRAND is available since Ivy Bridge CPU, so most of our platform should have it.

I think anyone should be able to try our MinBFT implementation in simulation mode without any problem.

nhoriguchi:
@sergefdrv sorry for my lack of words. My previous comment was intended to tell about how we fix seed generation *in* SGX SDK as you said. I *think* we have 2 options: using nanosecond timestamp as a seed, and using sgx_read_rand() as a seed instead of second timestamp.

>> and we might avoid additional library dependency as in nanosecond approach.
> I'm not sure what do you mean by additional library dependency.

sorry again. I just meant adding '#include <chrono>" may be needed, but actually that introduces no problem.

> I think anyone should be able to try our MinBFT implementation in simulation mode without any problem.

OK, so nanosecond approach is better.

> I was suggesting if some of us could prepare a patch for SGX SDK to use higher-precision timestamp for seeding the pseudo random number generation in simulation mode.

If you are OK, can I try this?

sergefdrv:
@Naoya-Horiguchi 

> I think we have 2 options: using nanosecond timestamp as a seed, and using sgx_read_rand() as a seed instead of second timestamp.

I'm not sure if `sgx_read_rand()` would be available at the point of enclave simulation initialization. `sgx_read_rand()` is provided by SGX SDK for enclave trusted code, whereas PRNG seed value is initialized in untrusted code. My suggestion is to change https://github.com/intel/linux-sgx/blob/54cae063cd0d21be5fab28c0d4b81b073b5d0914/sdk/simulation/urtssim/enclave_creator_sim.cpp#L232 to something using `clock_gettime()` with `CLOCK_MONOTONIC`. 

>>  I was suggesting if some of us could prepare a patch for SGX SDK to use higher-precision timestamp for seeding the pseudo random number generation in simulation mode.

> If you are OK, can I try this?

Please go ahead, I'll be focussed on implementing view change in MinBFT.

sergefdrv:
@Naoya-Horiguchi Would it be anything to change in this pull request, or can we merge it?

nhoriguchi:
@sergefdrv I'm fine to merge this series, thank you for your effort!

PR #18, opened 2018-08-06
Fix handling of UIs

sergefdrv:
Current implementation assumes that messages from a peer replica will be received and processed in sequence. The message handlers check and accept UI counter value before processing the message. That nevertheless can lead to a race condition when, after accepting UIs in sequence, the message processing and generation of new messages happens in arbitrary order. This can confuse other correct replicas. This series fixes the possible race condition by introducing two-state UI handling mechanism.

Current implementation also relies on the fact that all own messages are always handled by a replica as part of incoming message handling. That only applies to a primary replica so far. The primary handles its own Prepare messages by extracting it from corresponding Commit messages received from backup replicas. So the UI acceptor updates primary's own last accepted USIG counter value and the message gets processed. Since View Change will introduce more messages with UI, we cannot reply on that any more. Moreover, that is not a clean design solution. This series fixes handling of own messages with UI by eliminating processing own messages with a valid UI.


core/internal/peerstate/peerstate.go
```diff
@@ -0,0 +1,121 @@
...
+	committed       *sync.Cond
+}
+
+func (s *peerState) AcceptUI(ui *usig.UI) (new bool) {
+	s.Lock()
+	defer s.Unlock()
+
+	for ui.Counter-1 > s.lastAcceptedCV {
+		s.accepted.Wait()
+	}
```
sergefdrv, line 92: Actually, there is no need to wait for the previous counter value to be accepted. It is enough to wait for the previous counter value to be committed. I plan to update this commit in v2 of the PR, once the first round of review is finished.

luthlee: True, `s.lastAcceptedCV >= s.lastCommittedCV`.

luthlee (approved):

core/message-handling.go
```diff
@@ -38,33 +39,30 @@ type messageStreamHandler func(in <-chan []byte, reply chan<- []byte)
...
 func defaultMessageHandler(id uint32, log messagelog.MessageLog, config api.Configer, stack Stack) messageHandler {
 	n := config.N()
-	uiMessageChan := make(chan messages.MessageWithUI)
 
 	view := func() uint64 { return 0 } // view change is not implemented
 	clientStates := clientstate.NewProvider()
 	acceptUI := defaultUIAcceptor(stack)
 	collectCommit := defaultCommitCollector(id, clientStates, config, stack)
-	handleGeneratedUIMessages := defaultGeneratedUIMessageHandler(stack, log)
+	handleGeneratedUIMessage := defaultGeneratedUIMessageHandler(stack, log)
```
nhoriguchi: ~~Although it's a bit off-topic, I felt that local variables in this function might be named better (more readable), like acceptUI, collectCommit, handleGeneratedUIMessage, handleRequest, and handlePrepare. My preference is "uiAcceptor, commitCollector, generatedUIMessageHandler, requestHandler, and commitHandler." The reason is that they are all functions rather than objects (like UI, commit, message...).~~ It seems that the naming convention is like "acceptUI means a function to accept UI and acceptedUI is a UI that is accepted", so current names are consistent. Please ignore this paragraph.

And we can see the term "generated UI message" many times, does the word "generated" have special meaning? I mean that if you intend to use a separate counterpart like "non-generated" UI message somewhere, that's fine. But otherwise I thought we can go with only "UIMessage (meaning message with UI)" and "message (meaning message without UI)" .

core/internal/peerstate/peerstate.go
```diff
@@ -0,0 +1,121 @@
...
+// immediately, it will block until the UI can be accepted. The return
+// value new indicates if the UI has not been accepted before.
+//
+// CommitUI records the last accepted USIG unique identifier ui as
+// committed so that the next UI can be accepted. Each UI can be
+// committed one time only.
+type State interface {
+	AcceptUI(ui *usig.UI) (new bool)
+	CommitUI(ui *usig.UI) error
+}
```
sergefdrv, line 68: I'm actually I bit worried about the fact that the terms "accept" and "commit" may be too overloaded. Those also heavily used in MinBFT paper in various contexts like "accept message", "accept request", "COMMIT message". If we also count PBFT paper, that most of our users would be likely familiar with, it comes even worse. PBFT paper uses terms like "locally committed request", "committed request" etc.

So I just thought of possible change in terms to express the two-stage UI handling:

- `AcceptUI` → `CaptureUI`, or `AcquireUI`
- `CommitUI` → `ReleaseUI`, or `UnblockUI`, or `UnlockUI`

What do you think?

core/internal/peerstate/peerstate.go
```diff
@@ -0,0 +1,121 @@
...
+
+func (s *peerState) CommitUI(ui *usig.UI) error {
+	s.Lock()
+	defer s.Unlock()
+
+	if ui.Counter != s.lastAcceptedCV {
+		return fmt.Errorf("UI is not the last accepted")
+	} else if ui.Counter <= s.lastCommittedCV {
+		return fmt.Errorf("UI already committed")
+	}
```
nhoriguchi, line 115: Is it worth returning with error when `ui.Counter >= s.lastCommittedCV + 2`, which should never happen as long as AcceptUI/CommitUI work as intended?

core/prepare_test.go
```diff
@@ -46,57 +47,64 @@ func testMakePrepareHandlerPrimary(t *testing.T) {
...
 	mock.On("requestHandler", request, true).Return(replyChan, false, nil).Once()
 	new, err := handle(prepare)
 	assert.NoError(t, err)
 	assert.False(t, new, "UI already processed")
 
-	prepare = makePrepareMsg(view+1, primaryID(n, view))
-	mock.On("uiAcceptor", prepare).Return(true, nil).Once()
-	mock.On("uiCommitter", prepare).Once()
+	otherView := view + 1
+	otherPrimary := primaryID(n, view+1)
```
nhoriguchi: (... nitpicking) using `otherView` instead of `view+1`?

nhoriguchi (approved):
On the whole, this changeset looks good to me. So I agree to merge it (or next version.)

core/internal/peerstate/peerstate.go
nhoriguchi: I feel `CaptureUI` and `ReleaseUI` sounds good (matched well).

luthlee: Me too. I opt for `CaptureUI` and `ReleaseUI`

sergefdrv: :+1: Thanks, I'll change those names in v2.

core/internal/peerstate/peerstate.go
sergefdrv: I suppose you might wanted to ask: "Could it be that a replica somehow receives a valid UI  from another correct replica with the counter value `cv' > cv+1` before it receives the UI with the counter value `cv+1`?"

Consider the following situation. A backup replica receives a Prepare message directly from the primary with the counter value `cv` in UI. It starts handling the message by invoking `acceptUI` on the UI. `lastCommittedCV == cv-1` and `lastAcceptedCV == cv` at this point. Then, before the replica invokes `commitUI` on the UI, other two messages arrive from another backup replica: a Commit message for that Prepare message and a Commit message for the next Prepare message with the counter value `cv+1` in UI. Each of the Commit messages have the corresponding Prepare message embedded. Those embedded messages should be handled first. Note that message streams from different replicas are handled asynchronously. So the replica starts handling the embedded Prepare messages asynchronously with handling the Prepare message received directly from the primary. The first embedded Prepare has the counter value of `cv` in UI, so `acceptUI` simply returns `new == false`, and the message handling is finished. Then the first Commit message is handled normally. But the next embedded Prepare message would have the counter value `cv+1` in UI. Note that `lastCommittedCV == cv-1` at this point since `commitUI` hasn't yet been invoked for the UI with the counter value `cv`. So invoking `acceptUI` for the second embedded Prepare would satisfy the condition `ui.Counter >= s.lastCommittedCV + 2`. All the replicas were correctly following the protocol, so it should not be treated as error. Moreover, the replica should not drop the second Prepare, it needs to wait until the previous UI gets "committed" and then continue processing.

core/message-handling.go
sergefdrv: > And we can see the term "generated UI message" many times, does the word "generated" have special meaning? I mean that if you intend to use a separate counterpart like "non-generated" UI message somewhere, that's fine. But otherwise I thought we can go with only "UIMessage (meaning message with UI)" and "message (meaning message without UI)" .

The intention was to refer to a generated message that should have a UI assigned ("message with UI"). "Generated" was supposed to stress that the message has been just generated and has no UI assigned yet. 

As of identifier names in the code, `UIMessage` was a concise way to refer to "a message with UI (to be) assigned". Do you think it would make sense to change `GeneratedUIMessageHandler` and friends to something like `GeneratedMessageWithUIHandler`?

core/prepare_test.go
sergefdrv: :+1: Thanks!

core/internal/peerstate/peerstate.go
sergefdrv: > The first embedded Prepare has the counter value of `cv` in UI, so `acceptUI` simply returns `new == false`, and the message handling is finished.

Sorry, that is not quite true. `acceptUI` should actually wait until `commitUI` is invoked for the same UI.This is to be fixed: https://github.com/nec-blockchain/minbft/pull/19#discussion_r208379468. 

However, when more message types with UI are introduced, like Checkpoint, not every message with UI will always be embedded in messages generated by other replicas. Then it would be possible to receive a message from a correct replica that includes a message from another replica with UI counter value which is advanced more than by one. In that case, the replica should simply wait for the missing messages to be received directly, handle them, and resume handling of the embedded message.

It is great that you ask such kind of questions. It contributes to better understanding so that the implementation is going to be correct :wink:

PR #19, opened 2018-08-07
Fix handling of UIs v2

sergefdrv:
This is v2 of #18.

Summary of changes in v2:

- Terms "accept UI" and "commit UI" changed to "capture UI" and "release UI"
- Unnecessary conditional variable removed from `peerstate` package
- Comments to `uiCapturer` and `uiReleaser` (previously `uiAcceptor` and `uiCommitter`) refined

Please note, that comment refinement was not requested in code review.

PR #20, opened 2018-08-08
Fix handling of UIs v3

sergefdrv:
This is v3 of #19.

Summary of changes in v3:

- Added test to exercise capturing the same UI again before it released
- Fixed capturing the same UI
- Fixed test for handling Prepare for changed handling of own UIs
- Updated comments in peerstate package and `core/usig-ui.go`


core/internal/peerstate/peerstate_test.go
```diff
@@ -0,0 +1,178 @@
...
+	state := New()
+
+	wg.Add(nrConcurrent * nrUIs)
+	for workerID := 0; workerID < nrConcurrent; workerID++ {
+		workerID := workerID
+
+		go func() {
+			for cv := 1; cv <= nrUIs; cv++ {
+				assertMsg := fmt.Sprintf("Worker %d, UI %d", workerID, cv)
+				cv := cv
```
nhoriguchi, line 124: This short assignment seems trivial and has no effect?

nhoriguchi (approved):

core/internal/peerstate/peerstate_test.go
sergefdrv: TL;DR: This actually declares a new variable with the same name and value, which is used in the goroutine stated in the cycle. 

More detailed explanation of the issue: https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables.

PS. I believe some of the linters used by `gometalinter` will complain if this line is removed.

PR #24, opened 2018-08-09
Fix help of Makefile

ynamiki:
Target `protos` and `mocks` are described in the help but are not defined in the Makefile. This patch delete the non-existent targets from the help.


nhoriguchi (approved):
These seems obsolete targets.

sergefdrv (approved):
Nice catch :+1:

Issue #5, opened 2018-07-12
Consider faulty client


sergefdrv:
Faulty client might:
- not send a Request message to the primary
- send conflicting Request messages to different replicas

This could lead to expiration of the request timer and view change even if the primary is not faulty. This way liveness may be compromised.

sergefdrv:
Intuitively, #29 should fix the case 2, where the malicious client sends conflicting request IDs to different replicas. One option to fix the case 1 could be for a backup replica to forward client requests to the primary if there's no corresponding Prepare message within certain timeout, less than the request timeout.

nhoriguchi:
I think that we could introduce kind of "prepare timer" to detect case 1. But I'm not sure how replicas should respond to such events. Triggering view changes might not be the best option because that mean malicious clients easily/intentionally trigger view changes. If a client doesn't send (the same) request messages to *all* replicas, that request could be considered as "not finish submission process" so could be ignored silently. Maybe request timer also should trigger view changes only after confirming that the related request is properly submitted (f.e. complete the prepare phase?)

sergefdrv:
It seems that we cannot distinguish whether the client did not send the request to the primary, or the primary ignored it. I think the request timer should only initiate view change, if the replica is sure the primary had a chance to prepare the request. To that end, backup replicas might have to relay pending unprepared requests to the primary at some point. 

One approach is to relay the request upon the first expiration of the request timer, then restart it, and initiate view change if the request timer fires again. 

Another approach could be to introduce a dedicated timer, as you called it "prepare timer", that will trigger relaying the request to the primary. Once the request is relayed or prepared, the normal request timer should be started to initiate a view change if expired. 

I tend to slightly prefer the second approach. 

@jcs47 What do you think?

nhoriguchi:
> It seems that we cannot distinguish whether the client did not send the request to the primary, or the primary ignored it. I think the request timer should only initiate view change, if the replica is sure the primary had a chance to prepare the request. To that end, backup replicas might have to relay pending unprepared requests to the primary at some point.

OK, this is the precise description of the current problem.

> ...
> I tend to slightly prefer the second approach.

Thanks for the feedback. I'll write some code with the second approach. Maybe we might want to give shorter timeout value for "prepare timer" than the request timer, so that gives another reason for an additional timer.

PR #29, opened 2018-08-10
Fix request ID handling

sergefdrv:
This series combines fixing of various problems related to handling client request identifiers. These issues need to be solved to implement view change operation correctly. Summary of problems solved:
- Race condition when executing requests from different clients
- Handling of delayed reception of Request and Commit messages 
- Prepare/Commit message generation blocked by consensus rounds of previous requests
- Handling of duplicate Prepare messages and duplicate request execution attempts


core/internal/clientstate/client-state.go
```diff
@@ -156,10 +128,8 @@ func (c *clientState) AddReply(reply *messages.Reply) error {
 	c.Lock()
 	defer c.Unlock()
 
-	if seq != c.lastAcceptedSeq {
-		return fmt.Errorf("request ID mismatch")
-	} else if seq == c.lastExecutedSeq {
-		return fmt.Errorf("duplicate Reply")
+	if seq <= c.lastExecutedSeq {
```
nhoriguchi: Currently request seq number is generated based on timestamp. So if an attacker creates a request message on the system with wrong clock and large sequence number is used, any messages generated based on proper timestamp could be discarded as old requests?

nhoriguchi (approved):

core/internal/clientstate/client-state.go
sergefdrv: Please note that request identifiers (seq) are independent for each client. So if a compromised client exhausts its request ID space, it is screwed up. But only this client identity would be affected; others could continue normally. Although the consensus protocol cannot protect the client from being compromised, it provides safety and liveness when up to f replica nodes may be compromised (except tamperproof component).

nhoriguchi: OK, thank you for pointing out.

core/request.go
```diff
@@ -101,28 +97,18 @@ func makeRequestHandler(id, n uint32, view viewProvider, verifier messageSignatu
...
-		// replicas when received a new request from a client.
-		// It must probably be done atomically with checking
-		// the request ID, otherwise there might be a race
-		// condition between checking the request ID and
-		// starting the timer here versus stopping the timer
-		// and supplying a Reply message when the request gets
-		// executed.
+		// TODO: A new request ID has arrived; the request
+		// timer should be re-/started in backup replicas at
+		// this point.
```
ynamiki, line 133: Does this timer mean the timer for a view change described in page 6 in [Efficient Byzantine Fault-Tolerance paper](https://www.researchgate.net/publication/260585535_Efficient_Byzantine_Fault-Tolerance) and #5?

sergefdrv: Yes, that refers to the request timer which triggers the view change operation.

PR #30, opened 2018-08-27
Reorganize client state representation

sergefdrv:
Now the client state representation maintains two areas of functionality independently: handling of request identifiers and reply messages. Reorganize the code to reflect this split. This change is also desired to make it easier to extend the client state with request timeout handling. The request timeout is functionality required for view change.


nhoriguchi (approved):
looks good to me.

ynamiki (commented):
Looks good.

PR #32, opened 2018-10-01
CircleCI: Upgrade to SGX SDK 2.3.1

ynamiki:
As mentioned in #31, SGX SDK 2.3.1 includes a patch that can remove our workarounds. This PR upgrades the SDK used in the CI.

PR #34, opened 2018-10-05
README: Update Requirements to SGX SDK 2.3.1

ynamiki:
We have updated SGX SDK of the CI environment in #32.
Needs to update description about the testing environment in README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ export GOPATH="$HOME/go"
 ### Intel® SGX SDK ###
 
 The Intel® SGX enclave implementation has been tested with Intel® SGX
-SDK for Linux version 2.1.1. For installation instuctions please visit
+SDK for Linux version 2.3.1. For installation instuctions please visit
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -127,7 +127,7 @@ export GOPATH="$HOME/go"
 ### Intel® SGX SDK ###
 
 The Intel® SGX enclave implementation has been tested with Intel® SGX
-SDK for Linux version 2.1.1. For installation instuctions please visit
+SDK for Linux version 2.3.1. For installation instuctions please visit
 [installation guide][sgx-install]. A conventional directory to install
```
sergefdrv, line 131: Maybe also update the link to the installation guide. The link became broken. It seems to change quite often. Maybe this link would be more stable: https://01.org/intel-software-guard-extensions/downloads. Then it could be referred to as "download page" instead of "installation guide".

sergefdrv (changes requested):

PR #36, opened 2018-10-10
README: Update Requirements to SGX SDK 2.3.1

ynamiki:
We have updated SGX SDK of the CI environment in #32.
This patch updates description about the testing environment in README.


sergefdrv (approved):
One can say that this is rather a separate PR than a v2 of #34. So I just updated the PR title and description accordingly. I hope you don't mind.

PR #40, opened 2018-10-26
Split message handling

sergefdrv:
Currently, message handlers have message validation intermixed with message processing. However, in order to handle the `ViewChange` message, all messages embedded in it should also be validated. Any of those messages might be either `Prepare`, `Commit`, `ViewChange`, or `NewView`. Thus, a kind of recursive message validation is required. This prompts to restructure the message handling in such a way that message validation can be reused. Essentially, no-side-effect message validation should be split from further message processing. This message validation should fully verify message authenticity and consistency; it should have no side effect and should not depend of the current replica state.


client/message-handling.go
```diff
@@ -74,17 +78,36 @@ func startReplicaConnection(outHandler outgoingMessageHandler, inHandler incomin
...
+				panic(err)
+			}
+			out <- mBytes
+		}
 	}
 }
 
 // makeIncomingMessageHandler constructs an incomingMessageHandler
 // using the supplied request buffer as the destination and the
 // supplied authenticator for message authentication.
```
luthlee, line 94: You forget to update the comment accordingly

core/message-handling.go
```diff
@@ -40,15 +40,19 @@ type messageStreamHandler func(in <-chan []byte, reply chan<- []byte)
 // message hasn't been processed before.
 type messageHandler func(msg interface{}) (reply <-chan interface{}, new bool, err error)
 
+// messageConsumer receives a generated message.
```
luthlee, line 71: Based on this interface description, I cannot understand what kind of purpose does this interface serve and what is its relation to `messageHandler`.

Maybe you can update the description of `messageHandler` to something like "fully handles a **received** message" and rename it as `receivedMessageHandler`; and this interface as `generatedMessageRecorder` and the description as "generatedMessageRecorder records a generated message to update the replica's internal state. When handling a received message, a replica generates a message accordingly and deliveries it to the client or the other replicas. This is the last step before delivering the message to the other peers ."

core/usig-ui.go
```diff
@@ -31,17 +31,10 @@ import (
...
 // concurrently.
-type uiCapturer func(replicaID uint32, ui *usig.UI) (new bool)
-
-// uiReleaser synchronizes completion of sequential processing of USIG
-// unique identifier.
-//
-// It allows to continue sequential processing of UIs, synchronized by
-// uiCapturer. It is safe to invoke concurrently.
-type uiReleaser func(replicaID uint32, ui *usig.UI)
+type uiCapturer func(msg messages.MessageWithUI) (new bool, release func())
```
luthlee, line 37: Is `uiCapturer` going to be used for messages included in the view-change message? If so, then the comment of this interface should be updated for the following statement:
> Processing of UIs generated by the same replica 

Besides, why not to have a another function `parsedUiCapturer` to avoid parsing the UI message again?

client/message-handling.go
sergefdrv: Nice catch! Thanks!

core/message-handling.go
sergefdrv: I agree that `messageConsumer` might not be a good name for the thing. I would suggest naming it `generatedMessageConsumer` or `generatedMessageHandler` with a one-line summary like "takes care of generated message" (but a bit more precise). Maybe later we decide to move message signing and UI assigning in there. Then "recorder" might not be the best word to reflect the purpose. 

I would also suggest we rename `messageHandler`rather to `incomingMessageHandler`. This would be a bit more consistent with the naming in `client` package.

core/usig-ui.go
sergefdrv: No, I don't think `uiCapturer` would be useful to process the messages embedded into the ViewChange message.

If we were to reuse the UI parsed during message validation, we would need some way to pass it from `messageValidator` to `messageProcessor`. I think a better strategy would be to make the messages in `messages` package smarter. I was thinking of defining go structures for the messages that would be populated with all the pieces unserialized, like parsed UI, while also storing the original serialized stuff to check signatures etc. That would allow to parse the UI once the message is converted from serialized from to the go structure, and then simply reuse it. It would even allow to make the core independent from a particular mechanism of serialization, which currently happens to be protobuf.

core/message-handling.go
luthlee: Sounds good

Issue #42, opened 2018-10-30
Make some logging parameters configurable


nhoriguchi:
I learned from code review for #41 that peer/client can have more flexible logging by making some parameters (like log level, log file's path, and/or prefix format) configurable.

PR #44, opened 2018-11-22
peer/cmd: add logging-related command line options

nhoriguchi:
Here's a patch for issue #42. I commented that log prefix format can also be configurable, but it's not included finally because in my second thought default prefix string seems good enough.

### DIFF
--- a/core/utils.go
+++ b/core/utils.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/spf13/viper"
 	logging "github.com/op/go-logging"
 
 	"github.com/hyperledger-labs/minbft/api"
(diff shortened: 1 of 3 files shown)


core/utils.go
```diff
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/spf13/viper"
```
sergefdrv, line 24: I think we should avoid using this package in MinBFT core. It seems to be too specific for a particular target of integration. Maybe we should communicate logging parameters via optional parameters of the replica constructor?

sergefdrv (changes requested):

core/utils.go
nhoriguchi: @sergefdrv You're right, thanks. I'll update that part.

sergefdrv: Optional parameters could be passed as `opts ...Option` to `New()`. Then `Option` can be defined as `type Option func(*options)`, where `options` defined as `type options struct { /* some fields */ }`. Those parameters can be applied in `New()` like:
```go
	for _, opt := range opts {
		opt(&s.opts)
	}
```
And a concrete option can be implemented like:
```go
func WithLogFile(f *os.File) Option {
	return func(opts *options) {
		opts.logFile = f
	}
}
```
This way of implementing would be in line with what we already have prepared internally for view change implementation. Somewhat similar approach one can find used here https://godoc.org/github.com/micro/go-micro/client#NewStream

nhoriguchi:
This PR is updated with #45. So close it now.

PR #45, opened 2018-11-26
peer/cmd: add logging-related command line options (v2)

nhoriguchi:
Version 2 of #44 for original issue #42. The main change is to separate option parser code from core package by applying "functional options" pattern.

### DIFF
--- a/core/options.go
+++ b/core/options.go
@@ -1,0 +1,55 @@
+// Copyright © 2018 NEC Solution Innovators, Ltd.
+//
+// Authors: Naoya Horiguchi <n-horiguchi@ah.jp.nec.com>
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package minbft
+
+import (
+	"os"
+
+	logging "github.com/op/go-logging"
+)
+
+type Options struct {
+	LogLevel  logging.Level
+	LogFile   *os.File
+}
+
+type Option func(*Options)
+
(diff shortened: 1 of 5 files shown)


sample/peer/cmd/root.go
```diff
@@ -67,6 +67,14 @@ func init() {
...
 	must(viper.BindPFlag("keys",
 		rootCmd.PersistentFlags().Lookup("keys")))
+
+	rootCmd.PersistentFlags().String("logging-level", "", "logging level")
+	must(viper.BindPFlag("logging.level",
+		rootCmd.PersistentFlags().Lookup("logging-level")))
+
+	rootCmd.PersistentFlags().String("logging-file", "", "logging file")
+	must(viper.BindPFlag("logging.file",
+		rootCmd.PersistentFlags().Lookup("logging-file")))
```
sergefdrv, line 77: Adding this code to `root.go` would suggest that `peer request` command also supports those flags, whereas it doesn't. Please consider moving this code to `run.go`. Later on, we can add proper logging to the client module as well and support those flags for both `run` and `request` commands.

core/options.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"os"
+
+	logging "github.com/op/go-logging"
+)
+
+type Options struct {
+	LogLevel  logging.Level
+	LogFile   *os.File
+}
```
sergefdrv, line 28: Seemingly, there is no need to make this structure exported. Consider renaming it, as well as its fields, to start with low-case letters.

sergefdrv (changes requested):
Thanks a lot for your effort. I sill have some comments, but rather easy to address. This commit is relatively small, so it's okay to keep it all together. Otherwise it could also be split to (that would be too pedantic, though :laughing: ):
1. Add the mechanism to pass options when creating a replica instance
2. Support passing log options
3. Add CLI flags support to `peer` command

sample/peer/cmd/root.go
nhoriguchi: Right, I'll move this to run.go.

core/options.go
nhoriguchi: OK, thanks.

nhoriguchi:
Thank you for the comments. I'll submit the updated version soon.

nhoriguchi:
Updated with v3 (see #47).

Issue #16, opened 2018-07-27
CircleCI: Create and Use Custom Docker Image


ynamiki:
CircleCI, which we have set up in #13 supports custom Docker images so it is better to build a custom image with Intel SGX SDK and Go on Ubuntu 16.04 than to do everything in `.circleci/config.yml`.

A problem on using a custom image is that we need to have a Docker Registry or to manage an account of Docker Hub.

ynamiki:
To do:
- [Doing in the chat] Ask an administrator to install "Docker Hub Registry" application in hyperledger-labs organization of GitHub
- [Waiting the first item] Update configuration of Docker Hub to refer hyperledger-labs/minbft for automated build (ynamiki/minbft is used now)

ynamiki:
The above to-do items has been done. I have configured an autometed build from this repository.

ynamiki:
CircleCI was failed because it could not get the image. I will take a look.

https://circleci.com/gh/hyperledger-labs/minbft/tree/master

ynamiki:
The CI is now fixed (no changes, I just re-run the job); it seems that we need some time to get the latest image from Docker Hub.

nhoriguchi:
@ynamiki (just off-topic comment...) sorry for letting you close this issue manually. I added the reference "solved #16" in merge commit message, but according to https://help.github.com/articles/closing-issues-using-keywords/, the word "solved" is not a magic word. I'll use right word if I'll do merge operation next.

ynamiki:
@Naoya-Horiguchi No problem. It was my task to add the magic word in my commit comment.

PR #48, opened 2018-11-29
CircleCI: Use Custom Docker Image

ynamiki:
A Dockerfile is created for a custom image used in CircleCI and update configuration to use this (See #16).


nhoriguchi (commented):
Thanks for your work. Change itself looks good to me. Could you add some patch description for future reference? The subject line explains "what changes" well, so maybe explaining "why" or the effects/benefits in body will be helpful.

ynamiki:
@Naoya-Horiguchi Thanks for your comments. I have updated and opened the new issue #49.

PR #49, opened 2018-11-30
CircleCI: Use Custom Docker Image v2

ynamiki:
Currently, the CI sets up an environment (Go, Intel SGX SDK, etc.) from a fresh Ubuntu image everytime.

This patch introduces a custom Docker image, which contains everything used to build and test our project. By using this image, the CI can skip the environment setup and do just build and test in a CI process; saving resource and time.

Changes from v1 (#48):
- Reorganize commits
- Update the commit comment

### DIFF
--- a/.circleci/images/Dockerfile
+++ b/.circleci/images/Dockerfile
@@ -1,0 +1,26 @@
+FROM circleci/buildpack-deps:bionic-curl
+
+RUN sudo apt-get update \
+  && sudo apt-get install -y \
+    pkg-config
+
+RUN GO_URL="https://dl.google.com/go/go1.10.4.linux-amd64.tar.gz" \
+  && curl $GO_URL | sudo tar -C /usr/local -xzf - \
+  && echo 'export GOPATH="$HOME/go"' | sudo tee -a /etc/profile > /dev/null \
+  && echo 'export PATH="/usr/local/go/bin:$PATH"' | sudo tee -a /etc/profile > /dev/null
+
+RUN SGX_SDK_URL="https://download.01.org/intel-sgx/linux-2.3.1/ubuntu18.04/sgx_linux_x64_sdk_2.3.101.46683.bin" \
+  && sudo apt-get install -y \
+    build-essential \
+    python \
+  && curl $SGX_SDK_URL -o /tmp/sgx_linux_x64_sdk.bin \
+  && chmod +x /tmp/sgx_linux_x64_sdk.bin \
+  && echo -e "no\n/opt/intel" | sudo /tmp/sgx_linux_x64_sdk.bin \
+  && rm -f /tmp/sgx_linux_x64_sdk.bin \
+  && echo 'source /opt/intel/sgxsdk/environment' | sudo tee -a /etc/profile > /dev/null
+
+RUN GOMETALINTER_URL="https://github.com/alecthomas/gometalinter/releases/download/v2.0.11/gometalinter-2.0.11-linux-amd64.tar.gz" \
+  && curl -L $GOMETALINTER_URL | sudo tar -C /opt -xzf - \
+  && echo 'export PATH="/opt/gometalinter-2.0.11-linux-amd64:$PATH"' | sudo tee -a /etc/profile > /dev/null
+
+RUN echo 'export PATH="$GOPATH/bin:$PATH"' | sudo tee -a /etc/profile > /dev/null
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,39 +3,16 @@ jobs:
   build:
     working_directory: ~/go/src/github.com/hyperledger-labs/minbft
     docker:
-      - image: circleci/buildpack-deps:bionic
-    environment:
-      SGX_SDK: /opt/intel/sgxsdk
+      - image: necblockchain/minbft
     steps:
-      - run:
-          name: Install Go
-          command: |
-            wget "https://golang.org/dl/go1.10.3.linux-amd64.tar.gz" -O golang.tar.gz
-            sudo tar -C /usr/local -xaf golang.tar.gz
-            rm golang.tar.gz
-            echo 'export GOPATH=$HOME/go' >> $BASH_ENV
-            echo 'export PATH=$GOPATH/bin:/usr/local/go/bin:$PATH' >> $BASH_ENV
-      - run:
-          name: Install Intel SGX SDK
-          command: |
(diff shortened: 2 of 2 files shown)


nhoriguchi (approved):
Great, thank you!

sergefdrv:
@ynamiki Please wait a bit, I would have some comments.

.circleci/images/Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM circleci/buildpack-deps:bionic-curl
+
+RUN sudo apt-get update \
+  && sudo apt-get install -y \
+    pkg-config
+
+RUN GO_URL="https://dl.google.com/go/go1.10.4.linux-amd64.tar.gz" \
+  && curl $GO_URL | sudo tar -C /usr/local -xzf - \
+  && echo 'export GOPATH="$HOME/go"' | sudo tee -a /etc/profile > /dev/null \
+  && echo 'export PATH="/usr/local/go/bin:$PATH"' | sudo tee -a /etc/profile > /dev/null
```
sergefdrv, line 10: I was not sure why you use `| tee -a /etc/profile > /dev/null` rather than simply `>> /etc/profile`. Anyways, I would suggest adding this as a separate file in the line 9: `> /etc/profile.d/golang.sh`, then appending to the file with `>>` in the line 10.

.circleci/images/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+RUN SGX_SDK_URL="https://download.01.org/intel-sgx/linux-2.3.1/ubuntu18.04/sgx_linux_x64_sdk_2.3.101.46683.bin" \
+  && sudo apt-get install -y \
+    build-essential \
+    python \
+  && curl $SGX_SDK_URL -o /tmp/sgx_linux_x64_sdk.bin \
+  && chmod +x /tmp/sgx_linux_x64_sdk.bin \
+  && echo -e "no\n/opt/intel" | sudo /tmp/sgx_linux_x64_sdk.bin \
+  && rm -f /tmp/sgx_linux_x64_sdk.bin \
+  && echo 'source /opt/intel/sgxsdk/environment' | sudo tee -a /etc/profile > /dev/null
```
sergefdrv, line 20: Please also consider changing to `> /etc/profile.d/intel-sgx.sh`.

.circleci/images/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+RUN GOMETALINTER_URL="https://github.com/alecthomas/gometalinter/releases/download/v2.0.11/gometalinter-2.0.11-linux-amd64.tar.gz" \
+  && curl -L $GOMETALINTER_URL | sudo tar -C /opt -xzf - \
+  && echo 'export PATH="/opt/gometalinter-2.0.11-linux-amd64:$PATH"' | sudo tee -a /etc/profile > /dev/null
+
+RUN echo 'export PATH="$GOPATH/bin:$PATH"' | sudo tee -a /etc/profile > /dev/null
```
sergefdrv, line 26: Looks like this belongs to the `RUN` directive on line 7.

.circleci/images/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+    python \
+  && curl $SGX_SDK_URL -o /tmp/sgx_linux_x64_sdk.bin \
+  && chmod +x /tmp/sgx_linux_x64_sdk.bin \
+  && echo -e "no\n/opt/intel" | sudo /tmp/sgx_linux_x64_sdk.bin \
+  && rm -f /tmp/sgx_linux_x64_sdk.bin \
+  && echo 'source /opt/intel/sgxsdk/environment' | sudo tee -a /etc/profile > /dev/null
+
+RUN GOMETALINTER_URL="https://github.com/alecthomas/gometalinter/releases/download/v2.0.11/gometalinter-2.0.11-linux-amd64.tar.gz" \
+  && curl -L $GOMETALINTER_URL | sudo tar -C /opt -xzf - \
+  && echo 'export PATH="/opt/gometalinter-2.0.11-linux-amd64:$PATH"' | sudo tee -a /etc/profile > /dev/null
```
sergefdrv, line 24: Please also consider changing to `> /etc/profile.d/gometalinter.sh`.

.circleci/images/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+    build-essential \
+    python \
+  && curl $SGX_SDK_URL -o /tmp/sgx_linux_x64_sdk.bin \
+  && chmod +x /tmp/sgx_linux_x64_sdk.bin \
+  && echo -e "no\n/opt/intel" | sudo /tmp/sgx_linux_x64_sdk.bin \
+  && rm -f /tmp/sgx_linux_x64_sdk.bin \
+  && echo 'source /opt/intel/sgxsdk/environment' | sudo tee -a /etc/profile > /dev/null
+
+RUN GOMETALINTER_URL="https://github.com/alecthomas/gometalinter/releases/download/v2.0.11/gometalinter-2.0.11-linux-amd64.tar.gz" \
+  && curl -L $GOMETALINTER_URL | sudo tar -C /opt -xzf - \
```
sergefdrv, line 23: Please consider changing it like `tar --strip-components=1 -C /opt/gometalinter/`. Perhaps, we don't need to keep the version suffix in the path.

.circleci/images/Dockerfile
sergefdrv: I wonder if sourcing the `environment` file is enough? I remember I had to add another config file in `/etc/ld.so.conf.d/` on my machine so that the dynamic linker recognizes the SDK library path (LD_LIBRARY_PATH set in profile is ignored in recent Ubuntu, cf. https://github.com/hyperledger-labs/minbft#intel-sgx-sdk). One could check it in the container by trying one of the sample enclaves installed with SDK.

sergefdrv (changes requested):
Thanks!

.circleci/images/Dockerfile
sergefdrv: I see, `tee` is required to have access rights with `sudo`.

.circleci/images/Dockerfile
sergefdrv: This container apparently works if one specifies `-l` option to the shell. It's up to you, to either keep `-l` option or add loader config file.

ynamiki:
Thanks for review. I have updated and open a new PR.

PR #52, opened 2018-12-17
CI: Update Go Meta Linter v2.0.12

### LINKED ISSUE #51: Update gometalinter
Version 2.0.12 of `gometalinter` has been [released](https://github.com/alecthomas/gometalinter/releases/tag/v2.0.12). Let's update our CI image, too.

ynamiki:
Update gometalinter, which is used in the CI environment as a linter, to the latest release, v2.0.12.

This patch closes #51.

### DIFF
--- a/.circleci/images/Dockerfile
+++ b/.circleci/images/Dockerfile
@@ -18,7 +18,7 @@ RUN SGX_SDK_URL="https://download.01.org/intel-sgx/linux-2.3.1/ubuntu18.04/sgx_l
   && rm -f /tmp/sgx_linux_x64_sdk.bin \
   && sudo ln -s /opt/intel/sgxsdk/environment /etc/profile.d/intel-sgxsdk.sh
 
-RUN GOMETALINTER_URL="https://github.com/alecthomas/gometalinter/releases/download/v2.0.11/gometalinter-2.0.11-linux-amd64.tar.gz" \
+RUN GOMETALINTER_URL="https://github.com/alecthomas/gometalinter/releases/download/v2.0.12/gometalinter-2.0.12-linux-amd64.tar.gz" \
   && sudo mkdir /opt/gometalinter \
   && curl -L $GOMETALINTER_URL | sudo tar --strip-components=1 -C /opt/gometalinter -xzf -
 COPY gometalinter.sh /etc/profile.d


nhoriguchi:
@ynamiki, @sergefdrv  Thanks for the report/effort.

I found that GitHub provides the API to retrive the latest release info ([see here](https://developer.github.com/v3/repos/releases/#get-the-latest-release)). So you can get the latest tag like below (`jq` needs to be installed in the environment):

    curl -s https://api.github.com/repos/alecthomas/gometalinter/releases/latest | jq -r ".tag_name"
    v2.0.12

This can help us save effort to update the hard-coded line for every future updates.

ynamiki:
@Naoya-Horiguchi Thanks for the info. I think we can use this.
@sergefdrv I would like to hear your comments.

The docker container also has the following (version-specific) dependencies:

- Ubuntu
- Go
- Intel SGX SDK

I think I should leave them version-hard-coded since this project tightly depends on the environment.

On the other hand, gometalinter plays only a subsidiary role in this project. One problem of the method is any releases, even it has destructive changes, will be included into the image. But we can accept this and update the code to pass a new release of the linter soon.

sergefdrv:
@Naoya-Horiguchi This API to get GitHub release info seems attractive. I'm just not sure how it could trigger building of our CI docker image in Docker Hub. I'm curious, are you aware of any possible way to do so?

@ynamiki I agree that we should keep updating the core dependencies manually, because we refer to the versions in documentation. Otherwise, we would need to automatically update the readme file, too. Later on, when we would have better test coverage (thinking of more integration testing), we could rely on that to detect any compatibility issues.

We also have some vendored third-party Go packages. It would also be nice to update them regularly. I'd probably subscribe to those repos to receive release notifications...

sergefdrv (commented):
@ynamiki I wonder if CircleCI has used the updated docker image when testing this PR? At least, Docker Hub doesn't show any recent builds. It would be nice to catch any possible compatibility issues before we merge such kind of changes into master.

nhoriguchi:
Hi @sergefdrv,

> I'm just not sure how it could trigger building of our CI docker image in Docker Hub. I'm curious, are you aware of any possible way to do so?

We could use [automated build](https://docs.docker.com/docker-hub/builds/) of Docker Hub. I'll try to understand it more next.

nhoriguchi:
Automated build seems not our direct solution, because it's not designed to trigger building on push/release event of *other* repositories. [As answered here](https://stackoverflow.com/questions/40031793/trigger-automated-docker-hub-build-on-another-public-github-repo-change), we can effectively realize it with "Repository Link", which I think is to trigger building when the base image is updated. So one possible solution is that we add another Docker Hub repository to build intermediate layer of base image (which is rebuilt when some 3rd party repositrory is updated), and make necblockchain/minbft depend on it.

Anyway "intall latest gometalinter" part is independent of "automated build" part, so I'll make pull request for the former part later soon. For the latter part, any comments are welcomed.

ynamiki:
I had configured an automated build on Docker Hub as below:
- If code has pushed to `master`, build an image and tag with `latest`
- If code has pushed to another branch, an image is tagged with the branch name

https://cloud.docker.com/u/necblockchain/repository/docker/necblockchain/minbft/hubbuilds

A problem is I cannot find a way to start an automated build by creating a pull request. As Sergey mentioned, it would be nice if we make process like 1) create a PR, 2) build an image on Docker Hub, 3) run CI with the new image.  I'll welcome any comments.

nhoriguchi:
Hi @ynamiki san,

> I had configured an automated build on Docker Hub as below:
> 
>     * If code has pushed to `master`, build an image and tag with `latest`
> 
>     * If code has pushed to another branch, an image is tagged with the branch name

Thanks for the comment. We only have `master` branch now, so this configuration is fine. We will revisit this in the future when considering branching/tagging.

> A problem is I cannot find a way to start an automated build by creating a pull request.

I might miss something, but I think that maybe building is not supposed to run in every pull request. Or if you mean that we can run some logic like "check if the current latest image is uptodate and trigger building only if that's not true" at the beginning of CI process, that might be a possible solution.

ynamiki:
I am happy if it works as follows:

1. I open a new pull request
2. If `Dockerfile` is updated, Docker Hub build a new image automatically
3. CircleCI start a CI with the new Docker image

Currently, there is not step 2. So even I update `Dockerfile` as I did in this PR, CI uses an existing Docker image (not based on the updated Dockerfile).

nhoriguchi:
@ynamiki how about separating files related to buiding docker images into another git repository? That might be an easier solution (you don't have to write special logic to checking update).

ynamiki:
@Naoya-Horiguchi Thanks for your suggestion. I will try later.

I close this since we have #54 which overwrites this PR.

PR #53, opened 2018-12-17
Update vendored packages

sergefdrv:
First, the dependencies are split between those of the sample directory and the rest. Then, all the dependencies are updated to the latest release version of each directly imported package.


sergefdrv:
Somehow, CI check didn't run for this PR...

ynamiki:
> Somehow, CI check didn't run for this PR...

I am not sure why CircleCI did not work and it seems that there is no way to start a CI job manually....
I confirmed that this PR is built successfully in my environment.

nhoriguchi (approved):
@sergefdrv this is a big (~75k lines) change, so reviewers are not expected to read line-by-line. Could you share how you generated this change (maybe used some vendoring tool like `go dep`)? Separating patches for sample and others is nice.

ynamiki (approved):
Thanks.

[Go module](https://github.com/golang/go/wiki/Modules), which was experimentally introduced in Go 1.11 might provide better (and standard) way for managing dependencies. I will check it later.

sergefdrv:
@Naoya-Horiguchi 

> this is a big (~75k lines) change, so reviewers are not expected to read line-by-line. Could you share how you generated this change

You are right, this looks like a massive change, I should have provided some hint how to reproduce it. Nevertheless, the actual effect is recorded in the `vendor.json` files, that one can inspect. To reproduce the full change, one can remove all the files from the `vendor` directories, except `vendor.json`, and issue `govendor sync` command.

The first patch was a semi-manual move of parts of `vendor/vendor.json` to `sample/vendor/vendor.json`, followed by `govendor sync`. The second patch was achieved by removing all the vendored packages and adding the missing direct dependencies back, specifying the latest release version, with `govendor fetch name/of/the/package@version`.

sergefdrv:
@ynamiki 

> [Go module](https://github.com/golang/go/wiki/Modules), which was experimentally introduced in Go 1.11 might provide better (and standard) way for managing dependencies. I will check it later.

Thanks, that looks interesting. We might want to adopt this approach when it is stable.

sergefdrv:
I was a bit too fast merging this request. Now, `make lint` fails for some reason: https://circleci.com/gh/hyperledger-labs/minbft/77#tests/containers/0. Anyway, it seems like CircleCI silently ignores failures in linting stage. That might not be what we want.

sergefdrv:
> Now, `make lint` fails for some reason
> CircleCI silently ignores failures in linting stage

Fixed that in #55

PR #54, opened 2018-12-18
circleCI: build docker image with latest gometalinter release

nhoriguchi:
We can reduce hard-coding a version in Dockerfile, which will make code maintenance a bit easy.

Based on discussion in #52, and original issue is #51.

### DIFF
--- a/.circleci/images/Dockerfile
+++ b/.circleci/images/Dockerfile
@@ -18,7 +18,8 @@ RUN SGX_SDK_URL="https://download.01.org/intel-sgx/linux-2.3.1/ubuntu18.04/sgx_l
   && rm -f /tmp/sgx_linux_x64_sdk.bin \
   && sudo ln -s /opt/intel/sgxsdk/environment /etc/profile.d/intel-sgxsdk.sh
 
-RUN GOMETALINTER_URL="https://github.com/alecthomas/gometalinter/releases/download/v2.0.11/gometalinter-2.0.11-linux-amd64.tar.gz" \
-  && sudo mkdir /opt/gometalinter \
-  && curl -L $GOMETALINTER_URL | sudo tar --strip-components=1 -C /opt/gometalinter -xzf -
+COPY install_latest_gometalinter.sh /tmp/install_latest_gometalinter.sh
+RUN sudo apt-get install -y jq \
+  && bash /tmp/install_latest_gometalinter.sh
+
 COPY gometalinter.sh /etc/profile.d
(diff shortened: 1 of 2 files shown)


.circleci/images/Dockerfile
```diff
@@ -18,7 +18,8 @@ RUN SGX_SDK_URL="https://download.01.org/intel-sgx/linux-2.3.1/ubuntu18.04/sgx_l
   && rm -f /tmp/sgx_linux_x64_sdk.bin \
   && sudo ln -s /opt/intel/sgxsdk/environment /etc/profile.d/intel-sgxsdk.sh
 
-RUN GOMETALINTER_URL="https://github.com/alecthomas/gometalinter/releases/download/v2.0.11/gometalinter-2.0.11-linux-amd64.tar.gz" \
-  && sudo mkdir /opt/gometalinter \
-  && curl -L $GOMETALINTER_URL | sudo tar --strip-components=1 -C /opt/gometalinter -xzf -
+COPY install_latest_gometalinter.sh /tmp/install_latest_gometalinter.sh
+RUN sudo apt-get install -y jq \
+  && bash /tmp/install_latest_gometalinter.sh
```
ynamiki, line 23: Is it possible to write contents of the shell script here directly? I am wondering why there is a separate shell script.

.circleci/images/Dockerfile
```diff
@@ -18,7 +18,8 @@ RUN SGX_SDK_URL="https://download.01.org/intel-sgx/linux-2.3.1/ubuntu18.04/sgx_l
   && rm -f /tmp/sgx_linux_x64_sdk.bin \
   && sudo ln -s /opt/intel/sgxsdk/environment /etc/profile.d/intel-sgxsdk.sh
 
-RUN GOMETALINTER_URL="https://github.com/alecthomas/gometalinter/releases/download/v2.0.11/gometalinter-2.0.11-linux-amd64.tar.gz" \
-  && sudo mkdir /opt/gometalinter \
-  && curl -L $GOMETALINTER_URL | sudo tar --strip-components=1 -C /opt/gometalinter -xzf -
+COPY install_latest_gometalinter.sh /tmp/install_latest_gometalinter.sh
+RUN sudo apt-get install -y jq \
```
ynamiki, line 22:
```suggestion
RUN sudo apt-get install -y \
    jq \
```

.circleci/images/Dockerfile
nhoriguchi: Yes, firstly I considered that approach, but I felt it's hard to keep code readable when there's a line assigning variable from the output of other command. But anyway, I'll try to open it.

.circleci/images/Dockerfile
nhoriguchi: You're right, this is stated as [one of best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/).

nhoriguchi:
Updated version is uploaded, so close this issue.

PR #55, opened 2018-12-18
Do not ignore linting failures in CI

sergefdrv:
Currently, linting issues are not visible from PR check status because linting failures are ignored in CI. Make CI fail in case of any linting issue detected.

There are already some linting issues due to [the way `gometalinter` currently handles vendored packages](https://github.com/alecthomas/gometalinter/issues/449). Mask those explicitly with `--exclude` option.


nhoriguchi (approved):
Looks good to me.

ynamiki (approved):
Thanks!

PR #57, opened 2018-12-19
circleCI: build docker image with latest gometalinter release (v2)

nhoriguchi:
This is updated version of #54, with some refactoring applied. Original issue is #52.

PR #58, opened 2018-12-19
circleCI: build docker image with latest gometalinter release (v3)

nhoriguchi:
This is ver.3 of #57. The original issue is #52.

### DIFF
--- a/.circleci/images/Dockerfile
+++ b/.circleci/images/Dockerfile
@@ -18,7 +18,12 @@ RUN SGX_SDK_URL="https://download.01.org/intel-sgx/linux-2.3.1/ubuntu18.04/sgx_l
   && rm -f /tmp/sgx_linux_x64_sdk.bin \
   && sudo ln -s /opt/intel/sgxsdk/environment /etc/profile.d/intel-sgxsdk.sh
 
-RUN GOMETALINTER_URL="https://github.com/alecthomas/gometalinter/releases/download/v2.0.11/gometalinter-2.0.11-linux-amd64.tar.gz" \
+RUN sudo apt-get install -y \
+    jq \
+  && GOMETALINTER_VERSION="$(curl -s https://api.github.com/repos/alecthomas/gometalinter/releases/latest | jq -r '.tag_name')" \
+  && echo "Latest Gometalinter version: $GOMETALINTER_VERSION" \
+  && GOMETALINTER_URL="$(curl https://api.github.com/repos/alecthomas/gometalinter/releases/latest | jq -r '.assets[].browser_download_url | select(test("-linux-amd64.tar.gz$"))')" \
+  && echo "Downloading from $GOMETALINTER_URL" \
   && sudo mkdir /opt/gometalinter \
   && curl -L $GOMETALINTER_URL | sudo tar --strip-components=1 -C /opt/gometalinter -xzf -
 COPY gometalinter.sh /etc/profile.d


.circleci/images/Dockerfile
```diff
@@ -18,7 +18,12 @@ RUN SGX_SDK_URL="https://download.01.org/intel-sgx/linux-2.3.1/ubuntu18.04/sgx_l
   && rm -f /tmp/sgx_linux_x64_sdk.bin \
   && sudo ln -s /opt/intel/sgxsdk/environment /etc/profile.d/intel-sgxsdk.sh
 
-RUN GOMETALINTER_URL="https://github.com/alecthomas/gometalinter/releases/download/v2.0.11/gometalinter-2.0.11-linux-amd64.tar.gz" \
+RUN sudo apt-get install -y \
+    jq \
+  && GOMETALINTER_VERSION="$(curl -s https://api.github.com/repos/alecthomas/gometalinter/releases/latest | jq -r '.tag_name')" \
+  && echo "Latest Gometalinter version: $GOMETALINTER_VERSION" \
+  && GOMETALINTER_URL="$(curl https://api.github.com/repos/alecthomas/gometalinter/releases/latest | jq -r '.assets[].browser_download_url | select(test("-linux-amd64.tar.gz$"))')" \
```
sergefdrv, line 25: It would be nice to add `-s` option to the invocation of `curl`.

sergefdrv (commented):
Thanks for your effort. This change is indeed a good way to fetch the latest release when building the docker image. However, it is still not quite clear how to trigger rebuild of the image on Docker Hub, preferably without risking to break our CI. For instance, automated build can be triggered by merges into master branch. In that case, we might like updating this Dockerfile manually to trigger the build. So I would suggest to keep this PR open until we decide how to trigger the build.

nhoriguchi:
> So I would suggest to keep this PR open until we decide how to trigger the build.

OK, I agree that. We'll have discussion over this on #58.

ynamiki:
I found the official method to install the latest release of gometalinter:

`curl -L https://git.io/vp6lP | sh`

https://github.com/alecthomas/gometalinter#binary-releases

I can update the config file for CircleCI to use this after #60 is merged.

nhoriguchi:
@ynamiki ya, this could be a nice option. 👍

sergefdrv:
I'm not so sure that we should always use the latest version of the tool. The compiled binary of `gometalinter` and its bundled linters assume Go version they were compiled with: https://github.com/alecthomas/gometalinter/issues/449#issuecomment-448732522. So mismatching version of Go was the real reason why `gometalinter` unexpectedly reported issues in a vendor subdirectory, which was worked around in https://github.com/hyperledger-labs/minbft/pull/55/commits/dfe3cf702ee335f11b6915896013cfac8f3d216a. So we would need to eventually revert the workaround and make sure a compatible version of the tool is used. Moreover, if we fix the version then we would catch possible incompatibility _before_ merging version updates, thus keeping the master branch unaffected.

ynamiki:
Thanks for figure out the reason.

Then, we have options to revert the workaround for now. Note that our project is using Go 1.10 and the binary of gometalinter has been built with Go 1.10 (v2.0.6 or earlier) or 1.11 (v2.0.7 or later).

1. Hardcode a release of gometalinter in our CI configuration and use a compatible version of Go to build our project.
2. Every time in CI, download the latest gometalinter and build it along with our project

I would like to choose option 1 (i.e. use gometalinter 2.0.12 and upgrade to Go 1.11) to keep our built process simple.

sergefdrv:
@ynamiki I fully support your preference. But how did you figure out which version of Go each gometalinter release was built with?

ynamiki:
They are using Travis CI to make release archives. We can find the Go version from its configuration. The last line shows a binary compiled with Go 1.11 will be uploaded to the GitHub release page.

https://github.com/alecthomas/gometalinter/blob/master/.travis.yml

```
deploy:
  provider: releases
  api_key:
    secure: ...
  file_glob: true
  file:
    - dist/gometalinter-*.zip
    - dist/gometalinter-*.tar.bz2
    - dist/gometalinter-*.sha256
  skip_cleanup: true
  on:
    tags: true
    repo: alecthomas/gometalinter
    condition: $TRAVIS_GO_VERSION =~ ^1\.11
```

sergefdrv:
@ynamiki Thanks for sharing this :wink:

sergefdrv:
We don't use the docker image and decided to keep using a specific version of the tool.

PR #50, opened 2018-12-03
CircleCI: Use Custom Docker Image v3

ynamiki:
This is version 3 of #49.

Changes from v2:
- Setup environment variables in /etc/profile.d, not ~/.profile in a container
- Use COPY statement to create /etc/profile.d/*.sh
- Place gometalinter with version-independent directory: /opt/gometalinter

PR #60, opened 2018-12-25
CircleCI: Use Pre-Built Docker Image

### LINKED ISSUE #59: Check updates to CI Dockerfile
Since #50 has been merged, we use a custom docker image in CI, which is maintained on Docker Hub. The image is built from [Dockerfile](https://github.com/hyperledger-labs/minbft/blob/master/.circleci/images/Dockerfile) managed in the source tree. As long as this Dockerfile stays unchanged, it is safe to keep using the same image from Docker Hub when running CI checks. 

However, we would like to update the image from time to time. In that case, we should ensure the update will not break our CI. This comes down to running CI checks on the updated image before changing our default CI image on Docker Hub.

#52 attempts to update the Dockerfile without running CI checks on the updated image. #58 suggests changes to the Dockerfile to always fetch the latest release of `gometalinter` tool, but it's not clear how to trigger image build and ensure if the resulting image does not break CI. https://github.com/hyperledger-labs/minbft/pull/52#issuecomment-448104698 suggests using a separate repository for the Dockerfile.

We need to define a scheme for updating the CI docker image, preferably without risking to suddenly break our CI.

ynamiki:
By this patch, CircleCI uses a pre-built Docker image for CI.

Previously, we introduced a custom image (#50) with the aim of saving time and resources of a CI process. However, we are faced with problem that it is difficult to run a CI with a new Dockerfile when it is updated, as described in #59.

Since we found the saving of the time by a custom image is not so effective in our project as of now, we will go back to use a pre-built image.

This closes #59.


nhoriguchi (commented):
Thanks for working on this. This commit seems to do multiple things, reverting #50, introducing environment variables, using curl instead of wget, so it might be better to split commit for bisectability. And some command like curl and apt-get have silent options which will help us to have cleaner log output.

ynamiki:
@Naoya-Horiguchi Thanks for your comments.

For the reversion, I intended not physical (`git revert 50c86aa` (#50)) but logical (meaning just we use a pre-built image again) revert since the Dockerfile has been updated after #50; I tried to port the Dockerfile to CircleCI's config with saving its styles (using environment variables, `curl`, etc.).

Can I hear a little more about an appropriate approach for this? 
Should I do `git revert 50c86aa` (this cause conflict), then cherry-picking commits from #55 for this PR and create another PR for changing style (using environment variables, `curl`, etc.)?

> And some command like curl and apt-get have silent options which will help us to have cleaner log output.

I will do in a revised version of this PR.

nhoriguchi:
> For the reversion, I intended not physical (git revert 50c86aa (#50)) but logical (meaning just we use a pre-built image again) revert since the Dockerfile has been updated after #50; I tried to port the Dockerfile to CircleCI's config with saving its styles (using environment variables, curl, etc.).
>
> Can I hear a little more about an appropriate approach for this?
> Should I do git revert 50c86aa (this cause conflict), then cherry-picking commits from #55 for this PR and create another PR for changing style (using environment variables, curl, etc.)?

Ah OK, so you don't have to separate the commit (that makes things more complicated). Please keep going with current approach. Thank you for the explanation!

.circleci/config.yml
```diff
@@ -3,18 +3,41 @@ jobs:
...
+          name: Install Go
+          command: |
+            curl $GO_URL | sudo tar -C /usr/local -xzf -
+            echo 'export GOPATH=$HOME/go' >> $BASH_ENV
+            echo 'export PATH=$GOPATH/bin:/usr/local/go/bin:$PATH' >> $BASH_ENV
+      - run:
+          name: Install Intel SGX SDK
+          command: |
+            sudo apt-get update
+            sudo apt-get -y install build-essential python pkg-config
```
sergefdrv, line 23: Using `apt-get -qq <cmd>` would make the log a bit less noisy.

sergefdrv (changes requested):
Thanks for doing this! I think we can merge it after some small tweaks.

ynamiki:
Thanks for comments! I revised this in #61.

PR #62, opened 2019-01-03
Unify message processing

sergefdrv:
Current implementation of message processing has quite some duplication in code. Should we keep it as is, processing of view change messages would bring even more duplication. Thus, it is desirable to remove that duplication before. This pull request unifies message processing and eliminates the repeated logic.

First, for every message type, we introduce an abstraction to apply a message of that type to the current replica state. Then, those abstractions are used for handling of generated messages, as well. Finally, both processing of incoming message and handling of generated message are unified by using a generic abstraction to apply any replica message type.


sergefdrv:
NB. I had to force-push an identical commit series to trigger CI for some reason...

core/commit.go
```diff
@@ -160,7 +157,7 @@ func makeCommitCounter(f uint32) commitCounter {
 		defer lock.Unlock()
 
 		if prepareCV <= lastDoneCV {
-			return false, nil // ignore extra Commit
+			return true, nil // ignore extra Commit
```
nhoriguchi: Maybe you don't need the comment?

messages/messages.go
```diff
@@ -62,10 +67,11 @@ type MessageWithSignature interface {
...
 var (
 	_ ClientMessage        = (*Request)(nil)
-	_ MessageWithUI        = (*Prepare)(nil)
-	_ MessageWithUI        = (*Commit)(nil)
 	_ MessageWithSignature = (*Request)(nil)
+	_ ReplicaMessage       = (*Reply)(nil)
 	_ MessageWithSignature = (*Reply)(nil)
+	_ MessageWithUI        = (*Prepare)(nil)
+	_ MessageWithUI        = (*Commit)(nil)
 )
```
nhoriguchi, line 89: Sorry for golang question, what do these declarations actually do? These seem to me to do nothing effective (just assigning to blank identifiers)?

core/message-handling.go
```diff
@@ -127,12 +162,21 @@ func defaultIncomingMessageHandler(id uint32, log messagelog.MessageLog, config
 
 	applyCommit := makeCommitApplier(collectCommitment)
 	applyPrepare := makePrepareApplier(id, prepareSeq, collectCommitment, handleGeneratedUIMessage, applyCommit)
+	applyReplicaMessage := makeReplicaMessageApplier(applyPrepare, applyCommit)
 	applyRequest := makeRequestApplier(id, n, view, handleGeneratedUIMessage, applyPrepare)
 
+	var processMessage messageProcessor
+	processMessageThunk := func(msg interface{}) (new bool, err error) {
+		// delay evaluation of processMessage variable value
+		return processMessage(msg)
```
nhoriguchi, line 187: It's worth commenting why we need delayed evaluation here. Maybe to handle embedded messages?

core/message-handling.go
```diff
@@ -112,17 +112,27 @@ type replicaMessageApplier func(msg messages.ReplicaMessage) error
...
 // or the corresponding client, depending on the message type. The
 // message should be ready to serialize and deliver to the recipients.
-type generatedMessageHandler func(msg interface{})
-
-// generatedUIMessageHandler assigns and attaches a UI to a generated
-// message and arranges it to be delivered to peer replicas. It is
-// safe to invoke concurrently.
-type generatedUIMessageHandler func(msg messages.MessageWithUI)
+// It is safe to invoke concurrently.
+type generatedMessageConsumer func(msg messages.ReplicaMessage)
```
nhoriguchi, line 135: Do you have some reason to choose the word "consume"? In reviewing this PR, I have an impression that you introduce a few handlers with word "apply" or "applier", meaning "changing the state accordingly and producing any required side effects." So that looks to me fit here too.

nhoriguchi (commented):
I added some comments, but in total code/comments are well-written and series are well-constructed. And each commit looks to me have clear purpose. So I have no objection for merging it. I'll wait for other reviewers' feedback. Thanks for the work!

core/commit.go
sergefdrv: You seem to be right. Since we change the semantic of the returned value, it makes sense to remove this comment, as well.

messages/messages.go
sergefdrv: This is a kind of compile-time check that a type actually implements a certain interface.
No worries, It's fine to ask this kind of question :slightly_smiling_face:

core/message-handling.go
sergefdrv: This "thunk" is needed to construct a `applicableReplicaMessageProcessor` at line 175. If we were simply to pass `processMessage` there, that would be evaluated to `nil`, because it gets assigned later, at line 179. On the other hand, we also need to construct the `applicableReplicaMessageProcessor` before the `messageProcessor`, because the latter eventually depends on it. 

Basically, we have a recursion and a circular dependency here, and this has to be resolved somehow. Since Go uses strict evaluation (applicable order evaluation), we can resort to using this "thunk" and delay the evaluation of `processMessage` variable until it is actually used, after the whole recursive thing is constructed.

I need to think how to properly explain this in a comment...

core/message-handling.go
sergefdrv: I opted to use "consume", as opposed to "apply", because this "consumer" simply arranges messages to be delivered rather than manipulating any piece of the consensus state. So, "consume" here means to get the message and stash it wherever appropriate so that it can be delivered to the appropriate recipients, whenever they are ready to receive it.

sergefdrv: Moreover, all generated messages are applied by `generatedMessageHandler` to the local replica state. Reusing this term might add some confusion...

core/message-handling.go
nhoriguchi: > Basically, we have a recursion and a circular dependency here, and this has to be resolved somehow.

Yes, that's the point. I don't have a simple counter proposal of recursion-free approach, so I'm OK with current code.

core/message-handling.go
nhoriguchi: I understood the meaning and that's fine. Thank you for explanation.

ynamiki (approved):
Thanks, looks good to me.

ynamiki:
Are we right in assuming that we are waiting @luthlee's review?

sergefdrv:
Yes, and I'd also address the comments from @Naoya-Horiguchi when I have time

core/message-handling.go
sergefdrv: I'm thinking of adding the following comment to the code: _Due to recursive nature of message processing, an instance of messageProcessor is eventually required for it to be constructed. On the other hand, it is actually invoked only after getting fully constructed. This "thunk" delays evaluation of processMessage variable, thus resolving this circular dependency._

sergefdrv: ... and similar comment above declaration of `applyReplicaMessageThunk` in 8aac667572754dbfc3cfa33abb628be94d0cea37.

nhoriguchi: > 
> 
> I'm thinking of adding the following comment to the code: _Due to recursive nature of message processing, an instance of messageProcessor is eventually required for it to be constructed. On the other hand, it is actually invoked only after getting fully constructed. This "thunk" delays evaluation of processMessage variable, thus resolving this circular dependency._

Thank you. I like this comment.

PR #67, opened 2019-01-28
Sample replica connector tolerates f connection failure

nhoriguchi:
As stated in #1, current sample replica connector doesn't accept any connection failure to replicas, leading to whole system stall. This patchset tries to change this by making `peer-run`/`peer-request` commands fail to start when detecting more than `f` connection failure.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -59,7 +59,7 @@ func New(id uint32, n, f uint32, stack Stack) (Client, error) {
 
 	buf := requestbuffer.New()
 
-	if err := startReplicaConnections(id, n, buf, stack); err != nil {
+	if err := startReplicaConnections(id, n, f, buf, stack); err != nil {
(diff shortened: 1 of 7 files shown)


nhoriguchi:
I wrongly pushed a branch to hyperledger-labs/minbft, that's not our process. Really sorry. I closed this and will resubmit the PR in the right way.

sergefdrv:
@Naoya-Horiguchi You might find it useful to set up your git remote for this repository using the HTTPS URL (https://github.com/hyperledger-labs/minbft.git) instead of SSH (`git config remote.<name>.url <url>`). HTTPS URL doesn't allow to push, thus preventing from accidents. One can use `git push <URL> <local-branch-name>:<remote-branch-name>` when really needed to push without adding a permanent git remote.

Issue #65, opened 2019-01-25
Re-enable CircleCI checks in pull requests


sergefdrv:
CircleCI checks stopped functioning in pull requests recently, e.g. #64. We need to re-enable it.

ynamiki:
@sergefdrv It may sound strange, but I have not changed the configuration and my test PR (#66) triggered CircleCI without any problem.

[CircleCI will skip a build](https://circleci.com/docs/2.0/skip-build/) if `[ci skip]` or `[skip ci]` is included in a commit title or a description. But there are no such keywords in your commits.

I am sorry but I cannot identify a root cause. Could you please re-try a build by force-pushing into the branch if CircleCI is not triggered?

sergefdrv:
@ynamiki Lets keep an eye on it and see how it behaves next time. I remember I also had to do force-push in #62 to trigger CI.

I suspected that CI might not have been triggered because CircleCI was triggered against my private fork before I opened PR, but that should not be the case...

sergefdrv:
#70 did not trigger CI...

ynamiki:
I have no idea of the cause. The CI will not be triggered if a PR is opened by someone other than me...? I will look for related configuration.

ynamiki:
My PR #72 triggered the CI successfully🤔 (I have not changed any configuration).

sergefdrv:
#74 didn't trigger CI...

ynamiki:
I confirmed that GitHub triggers [the webhook](https://github.com/hyperledger-labs/minbft/settings/hooks/91605216) properly (e.g. `6cb75600-4159-11e9-8ca1-063fdf0a7fa5` for opening #75) and CircleCI replied 200 as a response. It seems that this is a matter of CircleCI (not GitHub).

sergefdrv:
Apparently, CircleCI Check's app has no access to code https://github.com/apps/circleci-checks/installations/718224. This might be the problem, as well...

sergefdrv:
Seems to have worked for #76.

ynamiki:
The CI has NOT triggered by #76 itself, but by merging it into `master`.
It seems that there are something wrong for a forked PR.

sergefdrv:
That's strange, I'm pretty sure I've seen CI check passed before I merged the #76. But I didn't go to the "Checks" tab; I looked at the panel on "Conversation" tab, just above "Merge" button. Let's keep it open and observe the behavior with the next PR.

[...]

PR #68, opened 2019-01-28
Sample replica connector tolerates f connection failures

nhoriguchi:
As stated in #1, current sample replica connector doesn't accept any connection failure to replicas, leading to whole system stall. This patchset tries to change this by making peer-run/peer-request commands fail to start when detecting more than `f` connection failure.

I updated some from #67 to fix test failures and gometalinter failures.

### DIFF
--- a/client/message-handling.go
+++ b/client/message-handling.go
@@ -30,17 +30,24 @@ import (
 // starts message exchange with them given a total number of replicas,
 // request buffer to add/fetch messages to/from and a stack of
 // interfaces to external modules.
-func startReplicaConnections(clientID, n uint32, buf *requestbuffer.T, stack Stack) error {
+func startReplicaConnections(clientID, n uint32, f uint32, buf *requestbuffer.T, stack Stack) error {
 	outHandler := makeOutgoingMessageHandler(buf)
 	authenticator := makeReplyAuthenticator(clientID, stack)
 	consumer := makeReplyConsumer(buf)
 	handleReply := makeReplyMessageHandler(consumer, authenticator)
 
+	faulty := uint32(0)
 	for i := uint32(0); i < n; i++ {
 		connector := makeReplicaConnector(i, stack)
 		inHandler := makeIncomingMessageHandler(i, handleReply)
 		if err := startReplicaConnection(outHandler, inHandler, connector); err != nil {
-			return fmt.Errorf("Error connecting to replica %d: %s", i, err)
+			faulty++
+			if faulty <= f {
+				fmt.Printf("Error connecting to replica %d: %s\n", i, err)
+				continue
+			} else {
+				return fmt.Errorf("Error getting connection failures from more than %d replicas", f)
+			}
 		}
 	}
 
--- a/core/replica.go
+++ b/core/replica.go
@@ -44,6 +44,7 @@ var _ api.MessageStreamHandler = (*Replica)(nil)
 type Replica struct {
 	id uint32 // replica ID, unique in range [0,n)
 	n  uint32 // total number of nodes
+	f  uint32 // number of tolerable faulty nodes
 
 	stack Stack
 
@@ -65,6 +66,7 @@ func New(id uint32, configer api.Configer, stack Stack, opts ...Option) (*Replic
 	replica := &Replica{
 		id: id,
 		n:  n,
+		f:  f,
 
 		stack: stack,
 
@@ -80,6 +82,7 @@ func New(id uint32, configer api.Configer, stack Stack, opts ...Option) (*Replic
 
 // Start begins message exchange with peer replicas
 func (r *Replica) Start() error {
+	f := uint32(0)
 	for i := uint32(0); i < r.n; i++ {
 		if i == r.id {
 			continue
@@ -87,7 +90,13 @@ func (r *Replica) Start() error {
 		out := make(chan []byte)
 		sh, err := r.stack.ReplicaMessageStreamHandler(i)
 		if err != nil {
-			return fmt.Errorf("Error getting peer replica %d message stream handler: %s", i, err)
+			f++
+			if f <= r.f {
+				fmt.Printf("Error getting peer replica %d message stream handler: %s\n", i, err)
+				continue
+			} else {
+				return fmt.Errorf("Error getting connection failures from more than %d peer replicas", r.f)
+			}
 		}
 		// Reply stream is not used for replica-to-replica
 		// communication, thus return value is ignored. Each
@@ -96,7 +105,13 @@ func (r *Replica) Start() error {
 		// connected.
 		_, err = sh.HandleMessageStream(out)
 		if err != nil {
-			return fmt.Errorf("Error establishing connection to peer replica %d: %s", i, err)
+			f++
+			if f <= r.f {
+				fmt.Printf("Error establishing connection to peer replica %d: %s\n", i, err)
+				continue
+			} else {
+				return fmt.Errorf("Error getting connection failures from more than %d peer replicas", r.f)
+			}
 		}
 
 		go func() {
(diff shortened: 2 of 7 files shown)


ynamiki:
I am not sure why CircleCI did not triggered.... I need to check again (see #65).

sample/peer/cmd/request.go
```diff
@@ -90,7 +92,9 @@ func request(req []byte) ([]byte, error) {
 
 	rc := connector.New()
 
-	err = rc.ConnectManyReplicas(peerAddrs, grpc.WithInsecure(), grpc.WithBlock())
+	ctx := context.Background()
+	ctx, _ = context.WithTimeout(ctx, 5*time.Second)
```
sergefdrv, line 96: It should be relatively easy to make this a configurable parameter (and add it to the sample config file). Nitpicking: It should also be fine to contract it to `ctx, _ := context.WithTimeout(context.Background(), timeout)`.

sample/net/grpc/grpc_test.go
```diff
@@ -24,6 +24,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/net/context"
```
sergefdrv, line 27: The package's documentation states:

> As of Go 1.7 this package is available in the standard library under the name context. 

So it is better to import it simply as `"context"` here and in other files.

client/message-handling.go
```diff
@@ -30,17 +30,24 @@ import (
...
 	consumer := makeReplyConsumer(buf)
 	handleReply := makeReplyMessageHandler(consumer, authenticator)
 
+	faulty := uint32(0)
 	for i := uint32(0); i < n; i++ {
 		connector := makeReplicaConnector(i, stack)
 		inHandler := makeIncomingMessageHandler(i, handleReply)
 		if err := startReplicaConnection(outHandler, inHandler, connector); err != nil {
-			return fmt.Errorf("Error connecting to replica %d: %s", i, err)
+			faulty++
```
sergefdrv, line 44: I'd suggest to rename it to `failed`.

core/replica.go
```diff
@@ -80,14 +82,21 @@ func New(id uint32, configer api.Configer, stack Stack, opts ...Option) (*Replic
 
 // Start begins message exchange with peer replicas
 func (r *Replica) Start() error {
+	f := uint32(0)
```
sergefdrv, line 85: I'd suggest to rename it to `failed`.

sample/peer/cmd/request.go
```diff
@@ -90,7 +92,9 @@ func request(req []byte) ([]byte, error) {
...
+	err = rc.ConnectManyReplicas(ctx, peerAddrs, grpc.WithInsecure(), grpc.WithBlock())
```
sergefdrv: Have you considered removing `grpc.WithBlock()` option? I tried to play a bit and replaced `grpc.WithBlock()` with `grpc.WithBackoffMaxDelay()`. Then added `grpc.FailFast(false)` to invocation of `h.client.Chat()` in `messageStreamHandler.HandleMessageStream()` of `sample/net/grpc/connector`, and wrapped most of its body in another goroutine. It seems to work... Maybe you could try experimenting a bit in this direction further?

sergefdrv (commented):
@Naoya-Horiguchi Thanks for working on this! I left some comments. Please look at the one regarding `grpc.WithBlock()` first: It might give an idea of how to solve the issue better using some features of gRPC.

sample/peer/cmd/request.go
sergefdrv: I haven't tried, but I think `grpc.WithBackoffMaxDelay()` may be even omitted for simplicity.

core/replica.go
```diff
@@ -96,7 +105,13 @@ func (r *Replica) Start() error {
 		// connected.
 		_, err = sh.HandleMessageStream(out)
 		if err != nil {
-			return fmt.Errorf("Error establishing connection to peer replica %d: %s", i, err)
+			f++
+			if f <= r.f {
+				fmt.Printf("Error establishing connection to peer replica %d: %s\n", i, err)
```
ynamiki, line 110: (Comment) Does a caller have a way to get how many/which nodes it can connect successfully? I guess a caller might want to know that programatically.

sample/net/grpc/connector/replica-connector.go
```diff
@@ -77,14 +77,13 @@ func (c *ReplicaConnector) ConnectReplica(ctx context.Context, replicaID uint32,
 
 // ConnectManyReplicas establishes a connection to many replicas given
 // a map from a replica ID to its gRPC target (address).
-func (c *ReplicaConnector) ConnectManyReplicas(ctx context.Context, targets map[uint32]string, dialOpts ...grpc.DialOption) error {
+func (c *ReplicaConnector) ConnectManyReplicas(ctx context.Context, targets map[uint32]string, dialOpts ...grpc.DialOption) {
 	for id, target := range targets {
 		err := c.ConnectReplica(ctx, id, target, dialOpts...)
 		if err != nil {
-			return err
+			fmt.Printf("Faled to connect peer %d: %s\n", id, err)
```
ynamiki, line 84: Faled -> Failed

ynamiki (changes requested):

sample/peer/cmd/request.go
nhoriguchi: > Have you considered removing `grpc.WithBlock()` option? I tried to play a bit and replaced `grpc.WithBlock()` with `grpc.WithBackoffMaxDelay()`. Then added `grpc.FailFast(false)` to invocation of `h.client.Chat()` in `messageStreamHandler.HandleMessageStream()` of `sample/net/grpc/connector`, and wrapped most of its body in another goroutine. It seems to work... Maybe you could try experimenting a bit in this direction further?

I didn't think much about `grpc.WithBlock()`, so OK, I'll try to do in your approach next.

sample/net/grpc/grpc_test.go
nhoriguchi: Right, so I fixed it.

client/message-handling.go
nhoriguchi: Applied, thanks.

core/replica.go
nhoriguchi: Applied, too.

sample/net/grpc/connector/replica-connector.go
nhoriguchi: Fixed. Thank you.

core/replica.go
nhoriguchi: Probably we can have it for example with `sync.WaitGroup`, I'l try it later.

sample/peer/cmd/request.go
sergefdrv: My experiment with removing `grpc.WithBlock()` and using `grpc.FailFast(false)` has shown that it seems to be possible to effectively mask connectivity failures solely by means of gRPC framework.

More importantly, this is also a matter of _our external API_. As MinBFT paper mentions that network faults are always masked, now I tend to think that we might need to refine the description of `ReplicaConnector` and `MessageStreamHandler` interfaces in `api/api.go`. Namely, require `ReplicaConnector.ReplicaMessageStreamHandler` to always succeed and return immediately, if the supplied parameter `replicaID` is correct. Additionally, remove error return value from `MessageStreamHandler.HandleMessageStream`, so that it is not allowed to fail.

Informally, those changes in API would mean that it should always be possible to immediately construct a local _representation_ of a remote replica. Moreover such a representation would be guaranteed to always be able to accept a channel of incoming messages and return a channel of messages produced in reply. In case of a network fault, the representation would simply stop sending/receiving messages to/from those channels, and then continue again, should the failure be resolved. Thus, the core would be free from handling of any kind of connectivity failure.

nhoriguchi:
The PR is updated with #69, so let's close this.

sample/peer/cmd/request.go
nhoriguchi: @sergefdrv hopefully the new version of PR is close to what you intended. I learned/tested some grpc options and finally think that I double-checked the suggested approach is quite better than v1. Thank you very much :smile:

PR #69, opened 2019-02-07
Sample replica connector tolerates f connection failures (v2)

nhoriguchi:
This PR is v2 of #68, but with quite different approach:
- by removing `grpc.WithBlock()` option, which makes `ConnectManyReplicas()` never fail
- by calling `ChannelClient#Chat()` with `grpc.FailFast(false)` *and* in goroutine, the main thread never gets stuck on connection failures. 

We still can't handle the cases like below:
- when the primary replica is unreachable,
- when some replica becomes unreachable after it once booted up properly.

These will be future works.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -66,7 +66,7 @@ type ReplicaConnector interface {
 // any of the channels, it is the receiver's responsibility to finish
 // handling of the message.
 type MessageStreamHandler interface {
-	HandleMessageStream(in <-chan []byte) (out <-chan []byte, err error)
+	HandleMessageStream(in <-chan []byte) (out <-chan []byte)
 }
 
 //======= Interface for module 'authentication' ========
--- a/api/api.go
+++ b/api/api.go
@@ -53,6 +53,8 @@ type Configer interface {
 // guarantees reliable and secure message delivery to the replica.
 // Message delivery delay is assumed not to grow indefinitely. This
 // assumption has to be satisfied to ensure the liveness of the system.
+// ReplicaMessageStreamHandler never fails except when incorrect replica
+// ID is passed.
 type ReplicaConnector interface {
 	ReplicaMessageStreamHandler(replicaID uint32) (MessageStreamHandler, error)
 }
@@ -64,7 +66,9 @@ type ReplicaConnector interface {
 // produced in reply. Each value sent/received through a channel is a
 // single complete serialized message. Once a message is received from
 // any of the channels, it is the receiver's responsibility to finish
-// handling of the message.
+// handling of the message. Another note is that the remote replica
+// might be unreachable, and in order to avoid whole system being stuck,
+// you should also initialize message stream asynchronously in goroutine.
 type MessageStreamHandler interface {
 	HandleMessageStream(in <-chan []byte) (out <-chan []byte)
 }
--- a/sample/net/grpc/connector/replica-connector.go
+++ b/sample/net/grpc/connector/replica-connector.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"sync"
 
 	"google.golang.org/grpc"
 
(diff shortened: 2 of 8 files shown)


api/api.go
```diff
@@ -64,7 +66,9 @@ type ReplicaConnector interface {
 // produced in reply. Each value sent/received through a channel is a
 // single complete serialized message. Once a message is received from
 // any of the channels, it is the receiver's responsibility to finish
-// handling of the message.
+// handling of the message. Another note is that the remote replica
+// might be unreachable, and in order to avoid whole system being stuck,
+// you should also initialize message stream asynchronously in goroutine.
```
sergefdrv, line 71: I would simply restate this like: "This method should never fail or block the caller."

sample/net/grpc/connector/replica-connector.go
```diff
@@ -97,14 +98,23 @@ type messageStreamHandler struct {
...
 	out := make(chan []byte)
 
-	stream, err := h.client.Chat(context.Background())
-	if err != nil {
-		return nil, err
-	}
+	wg.Add(1)
+	go func() {
+		var err error
+		stream, err = h.client.Chat(context.Background(), grpc.FailFast(false))
```
sergefdrv, line 108: We might want to update `grpc` package to v1.18.0 and use `grpc.WaitForReady` in favour of deprecated `grpc.FailFast`. You can use `govendor fetch google.golang.org/grpc/...@v1.18.0` from `sample` directory to do the update.

api/api.go
```diff
@@ -53,6 +53,8 @@ type Configer interface {
 // guarantees reliable and secure message delivery to the replica.
 // Message delivery delay is assumed not to grow indefinitely. This
 // assumption has to be satisfied to ensure the liveness of the system.
+// ReplicaMessageStreamHandler never fails except when incorrect replica
+// ID is passed.
```
sergefdrv, line 57: I believe, there should be a comma before 'except'. (Although, I'm totally not an expert in English, so I might be wrong.)

sergefdrv (changes requested):
I would elaborate a bit more in the commit message of 3e4ee168cf14d7d199079545dd4a3b99bd7bb49e as of why this is a good change to external API. You could say that, for a replica core or its remote representation (implementers of `MessageStreamHandler` interface), it should always be possible to accept a channel and return another channel. Moreover, according to the paper, network failures should always be masked, so this method should never return error.

After these changes, I'm pretty sure, we can make `Replica.Start` non-exported (lower-case it) and simply invoke it directly from `Replica.New`.

nhoriguchi:
> After these changes, I'm pretty sure, we can make `Replica.Start` non-exported (lower-case it) and simply invoke it directly from `Replica.New`.

I tried this and found that we can simply adjust `sample/peer/cmd/run.go` to the new semantics, but adjusting `core/integration_test.go` is not so easy. There is a dependency like below:

- `Replica` instance is available after `Replica.New`,
- we need `Replica` instance as `api.MessageStreamHandler` to set up dummy replica-connector,
- this setup needs to be done until calling `Replica.Start`.

If we merge `Replica.Start` into `Replica.New`, how can we give a `Replica` instance to replica-connector? I thought of defining in a package `connector` a type `messageStreamHandler` and its own method `messageStreamHandler.HandleMessageStream` which somehow calls `Replica.HandleMessageStream`, but I still don't make it. I start feeling that I want to separate this part as future work and move forward next version without that part.

All other comments are applied, thank you very much.

ynamiki (approved):
Thanks!

nhoriguchi:
@ynamiki thanks for your approval. I'll issue another PR later with grpc update and some comment refinement. As I stated, i'm still not sure about "private `Replica.Start`" part (still looking for a good fix).

ynamiki:
@Naoya-Horiguchi Sure, I understand the status.

sergefdrv:
> I tried this and found that we can simply adjust `sample/peer/cmd/run.go` to the new semantics, but adjusting `core/integration_test.go` is not so easy. 

I see the problem. Now, `sample/net/dummy/connector` doesn't always follow the changed connectivity API. Namely, it cannot provide a representation of another replica until its `ConnectReplica`/`ManyReplicas` method sets the corresponding replica instance.

> I thought of defining in a package `connector` a type `messageStreamHandler` and its own method `messageStreamHandler.HandleMessageStream` which somehow calls `Replica.HandleMessageStream`

:+1: I think this is the direction to possibly solve the issue. `messageStreamHandler.HandleMessageStream` would need to start a goroutine. That goroutine would wait for the target replica to be available, invoke that replica's `HandleMessageStream`, and start relaying messages of the reply channel. The hard part is waiting. One can close a channel (Go idiom) or use a `sync.Cond` (more traditional approach) to signal when the target replica instance becomes available.

> I start feeling that I want to separate this part as future work and move forward next version without that part.

I understand your doubts. But there is nothing apparently depending on this piece of work, so we can invest a little bit more time in it. More importantly, it would be better to clean up our external API earlier. Removing `Replica.Start` is prompted by the changes to connectivity API, so let's solve this little issue in testing. :wink:

nhoriguchi:
I updated the proposal with #70, so close this item.

PR #70, opened 2019-02-26
Sample replica connector tolerates f connection failures (v3)

nhoriguchi:
I've updated this series from v2 (#69),

- updated grpc to v1.18.0,
- improved comments,
- fixed `core/integration_test.go` (resultingly `sample/net/dummy/connector/replica-connector.go`)

I added many manual changes around `integration_test.go`, so you might want to comment on it...


sample/net/grpc/connector/replica-connector.go
```diff
@@ -97,14 +98,23 @@ type messageStreamHandler struct {
...
-	}
+	wg.Add(1)
+	go func() {
+		var err error
+		stream, err = h.client.Chat(context.Background(), grpc.WaitForReady(true))
+		if err != nil {
+			log.Printf("Error initializing client stream to replica %d: %s\n",
+				h.replicaID, err)
+		}
+		wg.Done()
```
ynamiki, line 113: `defer wg.Done()` should be placed at the beginning of the block?

sample/vendor/vendor.json
```diff
@@ -191,271 +237,299 @@
...
 			"revision": "51d6538a90f86fe93ac480b35f37b2be17fef232",
 			"revisionTime": "2018-11-15T11:05:04Z"
+		},
+		{
+			"path": "net/http/httptrace",
+			"revision": ""
 		}
 	],
-	"rootPath": "github.com/hyperledger-labs/minbft/sample"
+	"rootPath": "github.com/nec-blockchain/minbft/sample"
```
sergefdrv, line 534: Although `github.com/nec-blockchain/minbft` automatically redirects to `github.com/hyperledger-labs/minbft`, we'd better avoid using the old location.

sample/net/grpc/connector/replica-connector.go
```diff
@@ -97,14 +98,23 @@ type messageStreamHandler struct {
 // HandleMessageStream implements actual communication with remote
 // replica by means of the gRPC connection established to the replica.
 func (h *messageStreamHandler) HandleMessageStream(in <-chan []byte) (<-chan []byte, error) {
+	var stream proto.Channel_ChatClient
+	var wg sync.WaitGroup
```
sergefdrv, line 102: Maybe `sync.WaitGroup` is a bit overkill here. It is quite idiomatic in Go to close a channel for simple signalling like:
```go
ready := make(chan struct{})

go func() {
  /* ... */
  close(ready) // broadcast signal
  /* ... */
}()

/* ... */

<-ready // wait for signal
/* do stuff */
```

ynamiki (commented):
Thanks! I have just one comment.

sample/net/grpc/connector/replica-connector.go
nhoriguchi: OK, we can do this, so I'll update it :+1:

sample/net/dummy/connector/replica-connector.go
```diff
@@ -49,15 +49,53 @@ func (c *ReplicaConnector) ReplicaMessageStreamHandler(replicaID uint32) (api.Me
 	return replica, nil
 }
 
-// ConnectReplica assigns a replica instance to replica ID.
-func (c *ReplicaConnector) ConnectReplica(replicaID uint32, replica api.MessageStreamHandler) {
-	c.replicas[replicaID] = replica
+// ConnectReplica initializes a messageStreamHandler instance.
+func (c *ReplicaConnector) ConnectReplica(replicaID uint32, ready chan bool) {
+	c.replicas[replicaID] = &messageStreamHandler{replicaID, ready, nil}
```
sergefdrv, line 54: I think we don't have to change this function that way. Let me describe another approach.

We pass the total number of replicas to `New`. We add a map/slice of `chan bool` or `chan struct{}` in `ReplicaConnector` structure. The map/slice is indexed by replica ID. In `New`, we initialize this map/slice by making a channel for each replica ID. After we assign an instance of replica in `ReplicaConnector.ConnectReplica`, we also close the corresponding channel. This is a broadcast signal to all goroutines waiting in receive operation on that channel. (When a channel is closed, all receive operations on that channel return.) If we define `messageStreamHandler` as
```go
type messageStreamHandler struct {
	replicaID uint32
	connector *ReplicaConnector
}
```
then `ReplicaConnector.ReplicaMessageStreamHandler` could create and return an instance of `messageStreamHandler` by itself. `messageStreamHandler.HandleMessageStream` would wait until the corresponding channel is closed and use the replica instance assigned by `ReplicaConnector.ConnectReplica`.

sergefdrv (changes requested):
Sorry for delayed review. I left some comments. Thanks!

sample/vendor/vendor.json
nhoriguchi: My working directory under `$GOPATH` was still the old location. I fixed it.

sample/net/grpc/connector/replica-connector.go
nhoriguchi: Applied, thank you.

sample/net/dummy/connector/replica-connector.go
nhoriguchi: > ... `ReplicaConnector.ReplicaMessageStreamHandler` could create and return an instance of `messageStreamHandler` by itself.

This is what I missed in previous versions :astonished:.  In this approach, we can reduce the lines of changes in `core/integration_test.go`.

> We add a map/slice of `chan bool` or `chan struct{}` in `ReplicaConnector` structure.

Yes, relocating channels into `ReplicaConnector` looks fine to me, but I found that I got warnings from data race detector when doing like below:
<pre>
type ReplicaConnector struct {
    replicas map[uint32]api.MessageStreamHandler
    ready    map[uint32]chan bool
}

...

    c.replicas[replicaID] = replica
    c.ready[replicaID] <- true
}

...

func (h *messageStreamHandler) HandleMessageStream(in <-chan []byte) <-chan []byte {
    out := make(chan []byte)

    go func() {
        <-h.connector.ready[h.replicaID]
        replica := h.connector.replicas[h.replicaID]
        for msg := range replica.HandleMessageStream(in) {
            out <- msg
        }
    }()
</pre>
It seems to me that passing `Replica` instance via channel is better than "write-notify-wake-read" pattern, so I'm thinking of converting the type of `ReplicaConnector.replicas` from ` map[uint32]api.MessageStreamHandler` to `map[uint32]chan api.MessageStreamHandler`.

I have a next version of changeset with your approach, so will update PR soon.

sergefdrv: I think I know why you experienced data races in that code. Normally, a channel can act as a synchronization primitive. However in your case, the data race happened not because of concurrent access to the values of the keys, but due to the concurrent read/write access to the map itself. Since the total number of replicas are known at construction time, we can just allocate a slice of `api.MessageStreamHandler` and never change the slice itself afterwards. As opposed to maps, it is safe to access different indexes of the same slice concurrently.

nhoriguchi: Thank you for elaborating, I confirmed that I can fix it with using slice.

PR #72, opened 2019-03-02
Use Go Modules

### LINKED ISSUE #56: Use Go Modules
[Go 1.11 introduced "modules"](https://github.com/golang/go/wiki/Modules). This feature is now experimental and will be finalized in Go 1.12.

The feature may useful to manage dependencies. After Go 1.12 is released, we will consider whether it suits our project.

ynamiki:
Go 1.11 introduces a new concept called "modules," a build-in dependency management feature.
This patch makes changes to use the features, replacing govendor (so `vendor` directories are deleted).

This closes #56.


sergefdrv:
@ynamiki How did you convert to go modules? I tried to first remove the `vendor` directories and define all our source code as a single module. I've got somewhat more concise `go.mod` file:
```
module github.com/hyperledger-labs/minbft

go 1.12

require (
	github.com/a8m/envsubst v1.1.0
	github.com/golang/mock v1.2.0
	github.com/golang/protobuf v1.3.1
	github.com/inconshreveable/mousetrap v1.0.0 // indirect
	github.com/op/go-logging v0.0.0-20160315200505-970db520ece7
	github.com/spf13/cobra v0.0.3
	github.com/spf13/viper v1.3.2
	github.com/stretchr/objx v0.1.1 // indirect
	github.com/stretchr/testify v1.2.2
	golang.org/x/net v0.0.0-20190313220215-9f648a60d977
	golang.org/x/sync v0.0.0-20190227155943-e225da77a7e6
	google.golang.org/grpc v1.19.0
	gopkg.in/yaml.v2 v2.2.2
)
```

I am also not sure if and how we should keep this separation between core packages and those under `sample` directory. It is somewhat discouraged https://github.com/golang/go/wiki/Modules#faqs--multi-module-repositories.

ynamiki:
@sergefdrv As a first step, I do `go mod init` with `export GO111MODULE=on` in `minbft` and `minbft/sample` then remove `vendor` directories. I tried [conversion](https://github.com/golang/go/wiki/Modules#how-to-define-a-module) from `vendor.json` supported by `go mod init`. I have not checked whether result is reasonable and I cannot make it to be able to [build](https://circleci.com/gh/hyperledger-labs/minbft/100) at this time. 

Actually, I opened this (draft) PR to test the CI trigger (#65) so I have not studied the module system enough.

I will study the module system, including how should we organize modules in our project, then I will post some findings here. Thanks for providing the pointer about a multi-module project.

sergefdrv:
@ynamiki Thanks for clarifying this. We would also need to consider that `usig/sgx` is a special package because its `usig-enclave.go` file has to be generated with `make`, thus cannot be retrieved from the repository. Maybe we can change the way we build the enclave to make this an ordinary Go package.

Another concern would be regarding the assumption in the current code that the enclave binary is located at `$GOPATH/src/github.com/hyperledger-labs/minbft/usig/sgx/enclave/libusig.signed.so`. I can remember this assumption in peer command default configuration and somewhere in testing. After we have converted `minbft` into a Go module, it would no longer be an appropriate assumption.

sergefdrv:
> I cannot make it to be able to [build](https://circleci.com/gh/hyperledger-labs/minbft/100) at this time.

I think the problem is that Go requires `GO111MODULE=on` to be able to recognize a Go module under `$GOPATH/src`. Note that this is [not required](https://github.com/golang/go/wiki/Modules#how-to-install-and-activate-module-support) for directories outside of `$GOPATH/src`.

ynamiki:
@sergefdrv Some notes:

- Single module: We have two vendor directories (`minbft/vendor` and `minbft/sample/vendor`) before but we have one module now. I cannot find a strightforward way to define the two modules in our case. Go supports [a multi-module repository](https://github.com/golang/go/wiki/Modules#faqs--multi-module-repositories) and there is an example which is very similar to our case (we could have modules `.../minbft` and `.../minbft/sample` that share a prefix). Actually [I tried](https://github.com/ynamiki/minbft/commits/modularize-multi), but I faced [an error](https://circleci.com/gh/hyperledger-labs/minbft/120) in the test. I cannot fix this error by adding `require .../minbft/sample v0.0.0` and to `.../minbft/go.mod`. Since the multi-module style is somewhat [discouraged](https://github.com/golang/go/wiki/Modules#faqs--multi-module-repositories), this patch defines single module.
```
core/integration_test.go:31:2: unknown import path "github.com/hyperledger-labs/minbft/sample/authentication": cannot find module providing package github.com/hyperledger-labs/minbft/sample/authentication
```

- C codes: We have some codes written in C (i.e. `usig/sgx`). This means a module user needs to build this separately. Does Go system have something like hooks to build the codes automatically within `go get`?

nhoriguchi:
> * Single module: We have two vendor directories (`minbft/vendor` and `minbft/sample/vendor`) before but we have one module now. ...

Not a strong opinion, but generally speaking, I agree with single-module approach. The project is still small and we need focus on core functionality, so keeping other parts simple looks better to me. 

> * C codes: We have some codes written in C (i.e. `usig/sgx`). This means a module user needs to build this separately. Does Go system have something like hooks to build the codes automatically within `go get`?

According to [Go command page](https://golang.org/doc/articles/go_command.html#tmp_1), it seems that extra build configurations are decided not to be supported for simplicity. I'm not sure now how we work around this constraint, but it seems not "Go module" issue but "package delivery" issue. So I think we don't have to block this PR to be merged due to this issue.

nhoriguchi (commented):
Could you be more informative on patch subjects? We usually show list of commit titles to summarize changes when releasing, so better subjects mean better summary. 😉

ynamiki:
@Naoya-Horiguchi Thanks for your helpful comments. I have updated the subjects and move on to #82.

sergefdrv:
@ynamiki 

>     Single module

Given your rationale, I think it is fine to have a single module. At least, it is not an issue to focus on now.

>     C code

This is a bit more complicated matter. We can roughly split the code into three parts: 

- enclave code itself (shared library, C code)
- wrapper (shim) for the enclave (static library, C code)
- cgo wrapper around the shim library (part of Go codebase)

We cannot significantly change the way the enclave binary (`libusig.signed.so`) is build, if we keep using Intel SGX SDK.

The shim library was mainly introduced to hide low-level details of interaction with SGX SDK interface and the enclave instance itself. It was built as a static library rather than shared one because we didn't want to require users to install it in the system, neither use any potentially unsecure method to resolve dynamic linking dependencies, like setting `LD_LIBRARY_PATH` environment variable or passing `-rpath` option to the linker.

The cgo wrapper is generated by `make` from `usig-enclave.go.in` template file. This was necessary because it had to be linked with appropriate SGX SDK libraries: simulation- or hardware-based. The mode is determined by the user at build time. The cgo wrapper depends on the SDK libraries directly because the shim library is a static library and does not carry its own shared library dependency information.

As mentioned before, Go employs a simplistic approach to build packages. Namely, it only considers [certain file types](https://golang.org/cmd/go/#hdr-File_types) and simple built-in rules to compile those. This is certainly not enough to build the enclave binary, since it requires a special build process, including generation of edge routines. The generated edge routines are also required to compile the untrusted code - the shim library, in our case. So at least those parts have to be built separately from the main Go code base.

Another difficulty is the dependency on SDK dynamic libraries (sim/hw). We could either add those dependencies to cgo part, or make the shim library dynamic. We currently use the first option, thus require generated Go code. We could try to break the dependency by making the cgo part a dynamically loadable [Go plugin](https://golang.org/pkg/plugin/). But this would still require building Go code by unconventional building tools (`make`). Alternatively, we can make two identical cgo files, each depending on different set of SDK libraries (sim/hw). However, it is not clear how much code would have to be duplicated.

Making the shim library dynamic is another possible solution. Such a library can be easily built alongside with the enclave, automatically depending on correct SDK libraries. The issue of resolving the library itself could be solved by using `ldopen` in the main Go codebase. We can even embed the enclave binary into the shim library, thanks to the new SGX SGK API `sgx_create_enclave_from_buffer_ex`. This approach would allow to split building of the encalve-related C stuff from the main Go codebase. So one could build the main code base as a normal Go package. In order to run it with SGX USIG, one would also have to build the shim library, with enclave embedded. Then the path to the shim library could be specified as a configuration parameter.

ynamiki:
@sergefdrv Thank you for the details and your suggestions. The last option looks good for me, but I need to study security risk.

Can we proceed with #82, without changes in the build system, as a first step and separate the refinement as future work?

sergefdrv:
@ynamiki 
> Can we proceed with #82, without changes in the build system

I agree that we would benefit from Go modules. I plan to review #82 in the coming days.

PR #74, opened 2019-03-07
Sample replica connector tolerates f connection failures (v4)

nhoriguchi:
Update #70 with following changes:

- used defer statement for channel close,
- undid using `WaitGroup` for simple synchronization,
- fixed `rootPath` in `sample/vendor/vendor.json`,
- improved dummy `ReplicaConnector` and `core/integration_test.go`.


client/message-handling.go
```diff
@@ -119,7 +119,7 @@ func makeReplicaConnector(replicaID uint32, connector api.ReplicaConnector) repl
 		if err != nil {
 			return nil, fmt.Errorf("Error getting message stream handler: %s", err)
 		}
-		in, err := streamHandler.HandleMessageStream(out)
+		in := streamHandler.HandleMessageStream(out)
 		if err != nil {
 			return nil, fmt.Errorf("Error establishing connection: %s", err)
 		}
```
sergefdrv, line 125: Could you please also remove the orphaned check for error at line 123?

sample/net/dummy/connector/replica-connector.go
```diff
@@ -19,39 +19,35 @@
...
 
 // ReplicaMessageStreamHandler returns the instance previously
 // assigned to replica ID.
 func (c *ReplicaConnector) ReplicaMessageStreamHandler(replicaID uint32) (api.MessageStreamHandler, error) {
-	replica, ok := c.replicas[replicaID]
-	if !ok {
-		return nil, fmt.Errorf("No message stream handler registered for replica %d", replicaID)
-	}
-	return replica, nil
+	return &messageStreamHandler{replicaID, c}, nil
```
sergefdrv, line 45: Now, we need to check if `replicaID < uint32(len(c.replicas))`.

sample/net/dummy/connector/replica-connector.go
```diff
@@ -19,39 +19,35 @@
...
 }
 
 // ConnectReplica assigns a replica instance to replica ID.
 func (c *ReplicaConnector) ConnectReplica(replicaID uint32, replica api.MessageStreamHandler) {
-	c.replicas[replicaID] = replica
+	c.replicas[replicaID] <- replica
```
sergefdrv, line 50: This line will only wake a single goroutine created by `messageStreamHandler.HandleMessageStream` and waiting on the channel (or block, if there is no one), whereas the API doesn't forbid calling `MessageStreamHandler.HandleMessageStream` multiple times on the same instance. The replica representation should behave in this case as if it was a real replica instance: establish multiple independent connections. 

Closing a channel, on the other hand, is a non-blocking broadcast signal to all goroutines waiting in receive operation on that channel. This issue should be easy to solve by using pre-allocated slices of replica instances and ready channels. Please look at https://github.com/hyperledger-labs/minbft/pull/70#discussion_r263558017.

sergefdrv (changes requested):
Great job! I think we are nearly finished, just few more comments.

client/message-handling.go
nhoriguchi: Fixed it.

sample/net/dummy/connector/replica-connector.go
nhoriguchi: Right, it helps us know what happens for invalid input.

PR #75, opened 2019-03-08
Sample replica connector tolerates f connection failures (v5)

nhoriguchi:
I added more updates on #74:

- use slice of `api.MessageStreamHandler` instead of map of it in struct `ReplicaConnector` for better synchronization      
- add input check in `ReplicaMessageStreamHandler`
- remove orphaned error check in `makeReplicaConnector`


ynamiki (approved):

sample/net/dummy/connector/replica-connector.go
```diff
@@ -27,31 +27,37 @@ import (
...
+		return nil, fmt.Errorf("invalid replicaID")
 	}
-	return replica, nil
+	return &messageStreamHandler{replicaID, c}, nil
 }
 
 // ConnectReplica assigns a replica instance to replica ID.
 func (c *ReplicaConnector) ConnectReplica(replicaID uint32, replica api.MessageStreamHandler) {
 	c.replicas[replicaID] = replica
+	c.ready[replicaID] <- true
```
sergefdrv, line 60: We need to close the channel here, otherwise it's not a broadcast signal. Once the replica instance is ready, we want to unblock all goroutines waiting for it.

sergefdrv (changes requested):
Thank you, this PR has improved a lot! Just one last change, please :slightly_smiling_face:

sample/net/dummy/connector/replica-connector.go
nhoriguchi: Applied, thank you.

nhoriguchi:
Updated with #76, so I close this item.

PR #76, opened 2019-03-13
Sample replica connector tolerates f connection failures (v6)

nhoriguchi:
I updated from #75:

- closed channel in `ConnectReplica` to notifty readiness.


sergefdrv (approved):
Thanks!

PR #82, opened 2019-04-09
Use Go module (V2)

ynamiki:
This is version 2 of #72.

Changes from the previous version:
- Update the subjects (the first lines) of the commit messages


nhoriguchi (commented):
> Go module: delele `vendor` dir and create `go.mod` 

there is a typo: "delele" -> "delete".
This is a trivial issue, so you can do force-push to the proposing branch.

ynamiki:
@Naoya-Horiguchi Thanks, fixed.
> there is a typo: "delele" -> "delete".

README.md
```diff
@@ -156,9 +156,12 @@ export SGX_MODE=SIM
 
 ## Getting Started ##
 
-The project source code should be placed in
-`$GOPATH/src/github.com/hyperledger-labs/minbft/` directory. All
-following commands are supposed to be run from that directory.
+This is a Go module and can be placed anywhere; no need to be in
+GOPATH. If this is placed in GOPATH, please make sure that the
+environment variable `GO111MODULE=on` has set to activate module mode.
+
```
nhoriguchi, line 162: This seems to break the series of "getting started" steps. I tried to put `minbft` directory outside `$GOPATH` and unset the environment variable, then `bin/keytool generate` gets to fail with "Error: Failed to generate keys: failed to create USIG enclave: SGX error: 200f" message.

nhoriguchi: Ah, I missed that we still need to define GOPATH even with go module. Sorry for noise.

ynamiki: The environment value GOPATH is NOT required. Your error comes from absence of `libusig.signed.so`, which is built in the directory `minbft/usig/sgx/enclave`. `keytool` (and `peer`) depend on this and try to load it from `$GOPATH/src/github.com/hyperledger-labs/minbft/usig/sgx/enclave/libusig.signed.so` by default. If you put `minbft` outside GOPATH, it is required to specify a path for the library by a command-line option (`-u ../usig/sgx/enclave/libusig.signed.so`) etc., as described in the section "Running Example" of the README.

I think you might not give a path for the library to `keytool` and there might be another instance of the library in the default path (i.e. in `$GOPATH`).

nhoriguchi: Thanks for clarification, `-u` option is fine but <del>users typically try to build simply with `make build` so</del> we could have better default path. And there are a few more lines explicitly depending on `GOPATH` that could be affected by switching to Go module.

sergefdrv (approved):
LGTM, thanks! :+1:

README.md
nhoriguchi: I created #86 for this marginal issue.

PR #88, opened 2019-05-05
Remove Dependency on GOPATH

### LINKED ISSUE #86: settle explicit dependency on GOPATH
minbft is now treated as Go module with #82. Go module is free from GOPATH generally, but we still have dependency on it. Removing GOPATH is not mandatory or urgent but will be done at some point.

ynamiki:
`keytool` and `peer` commands load libusig.signed.so from `$GOPATH/src/github.com/hyperledger-labs/minbft/usig/sgx/enclave` by default. Since we are now a Go module and independent from GOPATH, we should rethink this default value.

This patch updates the default path to find the library from a current directory. This patch also update Makefile to add `install` target to install the commands and the library into a single directory (by default, `sample/bin` and `sample/lib` respectively), and sample configuration files to try this project with minimal configuration wherever this project is placed on a file system.

Note: variables in the Makefile are defined based on [Makefile conventions](https://www.gnu.org/software/make/manual/html_node/Makefile-Conventions.html#Makefile-Conventions) with some simplification.

This closes #86.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,2 @@
 *~
-bin
+/sample/build/
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-.PHONY: all build clean check test lint generate
+INSTALL := install
+INSTALL_PROGRAM := $(INSTALL)
+INSTALL_DATA := $(INSTALL) -m 644
+
+builddir := sample/build
+
+prefix := sample
+bindir := $(prefix)/bin
+libdir := $(prefix)/lib
+
+.PHONY: all build install uninstall clean check test lint generate
 
 all: build check
 test: check
@@ -31,6 +41,8 @@ help:
 	@echo 'Generic targets:'
 	@echo '  all (default)    - Build and test all'
 	@echo '  build            - Build all'
+	@echo '  install          - Install artifacts'
+	@echo '  uninstall        - Uninstall artifacts'
 	@echo '  clean            - Remove all build artifacts'
 	@echo '  check|test       - Run all tests'
 	@echo '  lint             - Run code quality checks'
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -14,7 +14,17 @@
...
 # limitations under the License.
 
-.PHONY: all build clean check test lint generate
+INSTALL := install
+INSTALL_PROGRAM := $(INSTALL)
+INSTALL_DATA := $(INSTALL) -m 644
+
+builddir := sample/build
+
+prefix := sample
```
sergefdrv, line 23: Would be better to wrap it into `ifdef prefix` to allow supplying a custom installation directory.

Makefile
```diff
@@ -40,12 +52,24 @@ help:
...
 
-build: usig-build | bin
-	go build -o sample/bin/peer ./sample/peer
-	go build -o sample/bin/keytool ./sample/authentication/keytool
+build: usig-build
+	go build -o $(builddir)/keytool ./sample/authentication/keytool
+	go build -o $(builddir)/peer ./sample/peer
+
+install: build
+	@mkdir -p $(bindir)
```
sergefdrv, line 60: It is recommended to use `install -d` in favor of `mkdir -p`. Alternatively, we can use `install -D` to copy files, which will create missing components of the destination file path.

.gitignore
```diff
@@ -1,2 +1,2 @@
 *~
-bin
+/sample/build/
```
sergefdrv, line 2: Please add `/sample/bin/` and `/sample/lib/`, as well.

sergefdrv (changes requested):

Makefile
ynamiki: I find [a conditional variable assignment operator `?=`](https://www.gnu.org/software/make/manual/html_node/Setting.html) (i.e.  `prefix ?= sample`), which seems to meet our requirements. Do you have any concerns about using this?

By using this, `make prefix=foo` installs the binaries in `foo/{bin,lib}`.

sergefdrv: You are right, `?=` would be much cleaner :+1:

ynamiki:
@sergefdrv Thanks for your comments. I have updated the code in #89.

PR #89, opened 2019-05-08
Remove Dependency on GOPATH (V2)

ynamiki:
This is version 2 of #88.

Changes from the previous version:

- Add the default installation directories to `.gitignore`
- Use `install -D` to install binaries with making target directories
- Now a user can change an installation directory with `make prefix=<dir>`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-.PHONY: all build clean check test lint generate
+INSTALL := install
+INSTALL_PROGRAM := $(INSTALL)
+INSTALL_DATA := $(INSTALL) -m 644
+
+builddir := sample/build
+
+prefix ?= sample
+bindir := $(prefix)/bin
+libdir := $(prefix)/lib
+
+.PHONY: all build install uninstall clean check test lint generate
 
 all: build check
 test: check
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -40,12 +52,22 @@ help:
...
 	@echo '                     $(usig-target-list)'
 
-build: usig-build | bin
-	go build -o sample/bin/peer ./sample/peer
-	go build -o sample/bin/keytool ./sample/authentication/keytool
+build: usig-build
+	go build -o $(builddir)/keytool ./sample/authentication/keytool
+	go build -o $(builddir)/peer ./sample/peer
+
+install: build
```
nhoriguchi, line 59: Personally I don't think that target `install` has to depend on target `build` because `make uninstall` had better be a reverse operation of `make install`. Or if you like to keep this dependency, maybe the help message of `make install` should be "Build and install artifacts".

sergefdrv: Although it might sound a bit unlogical, but it is actually a convention that `install` target builds first: https://www.gnu.org/prep/standards/html_node/Standard-Targets.html.

sergefdrv (approved):
Good job, thanks!

nhoriguchi (approved):

Makefile
ynamiki: I looked a Makefile generated by famous generators: [Automake](https://www.gnu.org/software/automake/), [CMake](https://www.gnu.org/software/automake/) and find that it follows the convention mentioned by @sergefdrv (CMake does not generate `uninstall` target [automatically](https://gitlab.kitware.com/cmake/community/wikis/FAQ#can-i-do-make-uninstall-with-cmake)). I would like to keep this code to follow the convention. Thanks for insightful review @Naoya-Horiguchi, and comments @sergefdrv.

PR #91, opened 2019-05-17
Add request timer

sergefdrv:
This series introduces request timer. The request timer is a mechanism to detect a faulty primary during normal case operation of the protocol. Expiration of a request timer will trigger the view change operation, when it is implemented.


ynamiki (approved):

core/internal/clientstate/request-timeout.go
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) 2018 NEC Laboratories Europe GmbH.
```
nhoriguchi, line 1: "2018-2019" or "2019"?

sergefdrv: This code was actually written in 2018. I'll update it to "2018-2019", then.

Issue #93, opened 2019-05-24
Requires libsgx_urts.so to run in SGX hardware mode


nhoriguchi:
Current installation procedure seems to cover only cases of running in SGX simulation mode. I saw the following error in running `keytool generate` with `SGX_MODE=HW`:

> bin/keytool: error while loading shared libraries: libsgx_urts.so: cannot open shared object file: No such file or directory

`libsgx_urts_sim.so` for simulation mode is installed by SGX SDK installer `sgx_linux_x64_sdk_2.3.101.46683.bin`, and `libsgx_urts.so` for hardware mode is delivered by `libsgx-enclave-common_2.3.101.46683-1_amd64.deb`. So we might be better to add some instruction for users who want to run MinBFT in hardware mode.

ynamiki:
Right, we need to install a driver and a platform software (PSW) in addition to the SDK to run the command in the hardware mode. We can update the README to ask a user to install these packages based on [the Intel's Installation Guide](https://download.01.org/intel-sgx/linux-2.3.1/docs/Intel_SGX_Installation_Guide_Linux_2.3.1_Open_Source.pdf).

@Naoya-Horiguchi Are you creating a patch for this? Shall I do this?

nhoriguchi:
@ynamiki, I'll work on this.

sergefdrv:
Please check if we actually need to install PSW. MinBFT doesn't seem to use any of PSW features. If I remember correctly, it was enough to install SGX driver prior to SDK to make hardware libraries available.

nhoriguchi:
@sergefdrv, thanks for the comment. Actually `libsgx_urts.so` is included in Intel SGX SDK, so installing PSW is not mandatory. It seems that the problem is that `libsgx_urts.so` and `libsgx_urts_sim.so` are located under different directories, and only `libsgx_urts_sim.so` is available under the library path `/opt/intel/sgxsdk/sdk_libs` (specified in `/etc/ld.so.conf.d/sgx-sdk.conf`):

> root@sgx3:~/tmp# find /opt/intel | grep urts
> /opt/intel/sgxsdk/include/sgx_urts.h
> /opt/intel/sgxsdk/lib64/libsgx_urts_sim.so
> /opt/intel/sgxsdk/lib64/libsgx_urts.so
> /opt/intel/sgxsdk/sdk_libs/libsgx_urts_sim.so
> /opt/intel/sgxsdk/pkgconfig/libsgx_urts_sim.pc
[...]

PR #95, opened 2019-06-03
README: update requirement for SGX hardware mode (v1)

nhoriguchi:
Based on #93, this PR suggests to update README to clarify the dependency on Intel SGX software components. I'm still not sure that this change makes full sense, so feel free to add comments and/or change requests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,20 +123,21 @@ please visit [this page](https://golang.org/doc/install).
 
 The Intel® SGX enclave implementation has been tested with Intel® SGX
 SDK for Linux version 2.3.1. For installation instuctions please visit
-[download page][sgx-downloads]. A conventional directory to install
+[download page][sgx-downloads].
+Please note that Intel SGX has two operation modes and required software
+components depend on operation mode.
+
+ - If we run in HW mode, we have to install all three components:
+   SGX driver, PSW, and SGX SDK.
+ - If we run in simulation mode, only SGX SDK is required.
+
+A conventional directory to install
 the SDK is `/opt/intel/`. Please do not forget to source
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -132,13 +140,6 @@ can add the following line to `~/.profile`:
 . /opt/intel/sgxsdk/environment
```
sergefdrv, line 140: As far as I can recall form my experience, sourcing `$SGX_SDK/environment` in `profile` was not enough. Modern Ubuntu system seems to reset `LD_LIBRARY_PATH` after processing commands from `profile`. That time I found `ld.so.conf` was the preferred way of telling the system to search for dynamic libraries in an additional directory. Could you please check that? If it is so, then let's just mention here that `sudo ldconfig` should be invoked once after adding the file.

nhoriguchi: > Modern Ubuntu system seems to reset LD_LIBRARY_PATH after processing commands from profile. That time I found ld.so.conf was the preferred way of telling the system to search for dynamic libraries in an additional directory.

You're right. I found [the related notice](https://help.ubuntu.com/community/EnvironmentVariables#File-location_related_variables), that says

> LD_LIBRARY_PATH: ... Since Ubuntu 9.04 Jaunty Jackalope, LD_LIBRARY_PATH cannot be set in $HOME/.profile, /etc/profile, nor /etc/environment files. You must use /etc/ld.so.conf.d/*.conf configuration files.

And ldconfig seems to be called within installation of PSW .deb package, so I'll add some to make it clear that `sudo ldconfig` is required for simulation mode.

Thanks!

PR #96, opened 2019-06-04
README: update installation and setup of Intel SGX SDK (v2)

nhoriguchi:
Based on feedback on v1 (#95), I updated the paragraph about `/etc/ld.so.conf.d/sgx-sdk.conf`.


sergefdrv (approved):
Thanks!

ynamiki (approved):
Thanks.

PR #97, opened 2019-06-05
Start request timer in primary

sergefdrv:
Even though any replica can always assume itself to be correct, a primary cannot safely assume its view to continue arbitrary long. Thus, it also needs to trigger view change in case some request times
out in its view.


nhoriguchi (approved):
looks good to me.

ynamiki (approved):

PR #98, opened 2019-06-07
Add view state component

sergefdrv:
This PR introduces a new package to track the state of view-change mechanism in a replica. The package is subsequently used through existing code.

### DIFF
--- a/core/internal/viewstate/view-state.go
+++ b/core/internal/viewstate/view-state.go
@@ -1,0 +1,272 @@
+// Copyright (c) 2018-2019 NEC Laboratories Europe GmbH.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package viewstate allows to manage current active view number. It
+// provides means for obtaining the current active view number, as
+// well as to coordinate the process of changing its value according
+// to the view-change mechanism.
+//
+// There are two distinct modes of operation for a replica: 'normal'
+// mode and 'view-change' mode.
+//
+// In the normal mode, a replica operates in a stable configuration,
+// called view, represented by the active view number. Initially, a
+// replica operates in the normal mode with an active view number
+// equal to zero. The active view can only be changed to a higher view
+// number via a transition through the view-change mode.
+//
+// A replica can request to change the view at any time. Requesting a
+// view change does not change the mode of operation between normal
+// and view-change, but helps peer replicas to coordinate the
+// beginning of a view-change operation.
+//
+// A replica begins a transition to a new view by first switching to
+// the view-change mode. Since there is no active view in this mode, a
+// replica waits for a new view to become the active one. It will then
+// shift to the normal mode as soon as there is a new established
+// active view. The new view number can be increased multiple times
+// before the view change is complete.
+//
+// Finally, as stated above, a replica transitions to the new view by
+// switching back to the normal mode when the view-change opreation is
+// complete.
+package viewstate
+
+import "sync"
+
+// State defines operations on the state of the view-change mechanism
+// maintained by each replica. All methods are safe to invoke
+// concurrently.
+//
+// WaitAndHoldActiveView waits for active view and defers view change.
+// If already in the view-change mode, it blocks and waits for the
+// normal mode. The returned view number is guaranteed to denote the
+// current active view until the returned release function is invoked.
+//
+// WaitAndHoldView waits for specific active view and defers view
+// change. This method resembles WaitAndHoldActiveView, but only waits
+// for a specific active view number supplied through the parameter
+// view. It immediately returns false once it is detected that the
+// supplied view number will never become active. Otherwise, it
+// returns true as soon as the awaited view number is active. In that
+// case, the view will stay active until the returned release function
+// is invoked.
+//
+// RequestViewChange handles synchronization of view change requests.
+// If the supplied view number is greater than the last requested and
+// the active view number then it returns true and records the view
+// number as last requested; otherwise, false is returned.
+//
+// StartViewChange synchronizes beginning of view change. If in the
+// normal mode and the supplied value is higher than the active view
+// number then it switches to the view-change mode with the expected
+// new view number equal to the supplied value. If already in the
+// view-change mode and the supplied value is higher than the expected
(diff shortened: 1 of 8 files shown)


core/internal/viewstate/view-state_test.go
```diff
@@ -0,0 +1,283 @@
...
+		assert.True(t, started <= view,
+			"Finished view change number cannot be lower than last started")
+		assert.True(t, active < view,
+			"Finished view change number must be higher than last active")
+		go release()
+	}
+
+	for i := 0; i < nrConcurrent; i++ {
+		for view := uint64(0); view <= nrViewChanges; view++ {
+			view := view
```
ynamiki, line 273: May I ask why we need this line? Is this a design pattern in concurrent processing?

core/internal/viewstate/view-state.go
```diff
@@ -0,0 +1,272 @@
...
+	lastRequestedView uint64
+
+	// Last started view change
+	lastNewView uint64
+
+	// Cond to signal on view change beginning/completion
+	viewChangeStartedFinished *sync.Cond
+
+	// Number of active view holders
+	nrActiveViewHolders int
```
ynamiki, line 122: I find both `nr` and `num` prefixes (like `numReplica` in other files) in variable names. Are there rules to distinguish these prefixes?

core/internal/viewstate/view-state.go
```diff
@@ -0,0 +1,272 @@
...
+//
+// A replica begins a transition to a new view by first switching to
+// the view-change mode. Since there is no active view in this mode, a
+// replica waits for a new view to become the active one. It will then
+// shift to the normal mode as soon as there is a new established
+// active view. The new view number can be increased multiple times
+// before the view change is complete.
+//
+// Finally, as stated above, a replica transitions to the new view by
+// switching back to the normal mode when the view-change opreation is
```
nhoriguchi, line 42: (sorry for nitpicking...) s/opreation/operation/

core/internal/viewstate/view-state.go
```diff
@@ -0,0 +1,272 @@
...
+type State interface {
+	WaitAndHoldActiveView() (view uint64, release func())
+	WaitAndHoldView(view uint64) (ok bool, release func())
+	RequestViewChange(newView uint64) bool
+	StartViewChange(newView uint64) (ok bool, release func())
+	FinishViewChange(newView uint64) (ok bool, release func())
+}
+
+// In the normal mode, if lastNewView == lastActiveView. Otherwise, in
+// the view-change mode, and lastNewView > lastActiveView.
```
nhoriguchi, line 105: This comment is very helpful, and it would be more readable if you introduce some helper functions named (for example) `isNormalMode`/`isViewChangeMode`. Checking state conditions in raw numbers are sometimes hard to understand what the conditions mean.

core/message-handling.go
```diff
@@ -157,13 +157,16 @@ func defaultIncomingMessageHandler(id uint32, log messagelog.MessageLog, config
...
+	viewState := viewstate.New()
 
 	captureSeq := makeRequestSeqCapturer(clientStates)
 	prepareSeq := makeRequestSeqPreparer(clientStates)
 	retireSeq := makeRequestSeqRetirer(clientStates)
 	startReqTimer := makeRequestTimerStarter(clientStates, handleReqTimeout, logger)
 	stopReqTimer := makeRequestTimerStopper(clientStates)
 	captureUI := makeUICapturer(peerStates)
+	provideView := viewState.WaitAndHoldActiveView
+	waitView := viewState.WaitAndHoldView
```
nhoriguchi, line 169: It sounds to me surprising to see `WaitAndHoldActiveView` is used to get current view. But I can't clearly say what is better, but `WaitAndHoldActiveView` could be named better?

sergefdrv: Here, `provideView` just allows to get the current active view. But whenever the current active view is required, it can only be acquired when it becomes available, i.e. in normal mode only. Once the view number is acquired, it must be guaranteed not to change, thus has to be released afterwards. The name `WaitAndHoldActiveView` attempts to reflect this logic.

The name `viewProvider` is more concise and generic, since the code using it (`makeRequestApplier`) doesn't really care about those details - it only has to obey the requirement to release any acquired view. In contrast, the code supplying this abstraction (`defaultIncomingMessageHandler`) has to ensure that this primitive is used consistently with the other things, thus it has to deal with `viewstate` package and consider all those details.

We could also introduce trivial functions `makeViewProvider`, `makeViewWaiter`, that will take `viewState` as a parameter. But since `WaitAndHoldActiveView`/`WaitAndHoldView` matches `viewProvider`/`viewWaiter` in definition, it can be simply used directly.

core/internal/viewstate/view-state.go
sergefdrv: Thank you for your suggestion. Let me think of some helper function to make the code read easy. So far I can see that `isViewChangeMode` would not be used anywhere. But other kind of helper might be useful.

ynamiki (commented):
No request for changes, but I have some questions.

core/internal/viewstate/view-state_test.go
sergefdrv: This is a common idiom in Go. It is typically used when using a function literal (lambda/anonymous function) to start a goroutine from a loop. This line creates a new independent variable `view` for each iteration of the loop. The new variable captures the current value of the loop variable `view` and shadows it. Then any following goroutine started from in the loop will use that new variable. Otherwise, the started goroutine would use the original variable that keep asynchronously changing as the loop continues to iterate.

core/internal/viewstate/view-state.go
sergefdrv: It seems `num` prefix is only used in `integration_test.go`. I think both prefixes mean the same thing, it was just a matter of personal preference.

core/internal/viewstate/view-state.go
sergefdrv: What I can suggest are the following helpers.

```go
func (s *viewState) inNormalMode() bool {
	return s.lastActiveView == s.lastNewView
}

func (s *viewState) inView(view uint64) bool {
	return s.inNormalMode() && view == s.lastActiveView
}

func (s *viewState) isNewView(view uint64) bool {
	return view > s.lastActiveView
}

func (s *viewState) isNewViewCandidate(view uint64) bool {
	return view > s.lastNewView
}

func (s *viewState) isPossibleNewView(view uint64) bool {
	return s.isNewView(view) && view >= s.lastNewView
}
```

[Here](https://github.com/sergefdrv/minbft/compare/add-view-state..sergefdrv:add-view-state-2) is how the code would change. Does it read better?

nhoriguchi: Very nice, thank you very much.

core/internal/viewstate/view-state_test.go
ynamiki: Thank you for explanation. I understand.

I find that this is FAQ. Sorry to bother you.
https://golang.org/doc/faq#closures_and_goroutines

Issue #28, opened 2018-08-09
Summarize Contributing Rules


ynamiki:
Summarize the rules described in #12 in https://github.com/hyperledger-labs/minbft#contributing or in separate file as suggested at https://help.github.com/articles/setting-guidelines-for-repository-contributors/.

PR #101, opened 2019-07-11
Generic replica connectivity

jcs47:
This pull request aims to make the replica connectivity more generic. The idea is to allow other projects to re-use the connectivity while having the liberty to supply their own ordering protocol.  For example, one could want to implement another protocol that isn't MinBFT but want to use the connectivity implemented in this project.

To this end, the functions `incomingMessageHander` and `replyCollector` were replaced by interfaces `ProtocolHandler` and `ReplyCollector`, respectivly. Moreover, there are also the following new functions:

```
func GetReplica(id uint32, configer api.Configer, stack Stack, handler api.ProtocolHandler, logger *logging.Logger) (*Replica, error)
func GetClient(id uint32, n, f uint32, stack Stack, collector ReplyCollector) (Client, error)
func GetMinBFTHandler(id uint32, log messagelog.MessageLog, config api.Configer, stack Stack, logger *logging.Logger) api.ProtocolHandler
func GetLogger(id uint32, opts ...Option) *logging.Logger
func GetMinBFTCollector(f uint32) ReplyCollector
```

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -23,13 +23,20 @@ import (
 )
 
 // ProtocolHandler handles incoming message according to the implemented protocol.
-//
-// Handle verifies and processes incoming messages according to the protocol.
-// If there is any message produced in reply, it will be send to reply
-// channel, otherwise nil channel is returned. The return value new
-// indicates that the message has not been processed before.
 type ProtocolHandler interface {
+
+	// Handle verifies and processes incoming messages according to the protocol.
+	// If there is any message produced in reply, it will be send to reply
+	// channel, otherwise nil channel is returned. The return value new
+	// indicates that the message has not been processed before.
 	Handle(msg interface{}) (reply <-chan interface{}, new bool, err error)
+
+	Wrap(msg interface{}) (msgBytes []byte, err error)
+
+	// Unwrap deserializes messages into an interface compatible with the ordering
+	// protocol, alongside a string describing the message type. An error is
+	// returned if deserialization fails.
+	Unwrap(msgBytes []byte) (msg interface{}, msgStr string, err error)
 }
 
 //======= Interface for module 'config' =======
--- a/api/api.go
+++ b/api/api.go
@@ -31,9 +31,11 @@ type ProtocolHandler interface {
 	// indicates that the message has not been processed before.
 	Handle(msg interface{}) (reply <-chan interface{}, new bool, err error)
 
+	// Wrap serializes a message into a byte array. An error is returned if the
+	// serialization failed.
 	Wrap(msg interface{}) (msgBytes []byte, err error)
 
-	// Unwrap deserializes messages into an interface compatible with the ordering
+	// Unwrap deserializes a message into an interface compatible with the ordering
 	// protocol, alongside a string describing the message type. An error is
 	// returned if deserialization fails.
 	Unwrap(msgBytes []byte) (msg interface{}, msgStr string, err error)
--- a/api/api.go
+++ b/api/api.go
@@ -22,6 +22,16 @@ import (
 	"time"
 )
 
+// ProtocolHandler handles incoming message according to the implemented protocol.
+//
+// Handle verifies and processes incoming messages according to the protocol.
+// If there is any message produced in reply, it will be send to reply
+// channel, otherwise nil channel is returned. The return value new
+// indicates that the message has not been processed before.
+type ProtocolHandler interface {
+	Handle(msg interface{}) (reply <-chan interface{}, new bool, err error)
+}
+
 //======= Interface for module 'config' =======
 
 // Configer defines the interface to obtain the protocol parameters from
--- a/api/api.go
+++ b/api/api.go
@@ -25,6 +25,9 @@ import (
 // ProtocolHandler handles incoming message according to the implemented protocol.
 type ProtocolHandler interface {
 
+	// Start begins message exchange with peers.
+	Start() error
+
 	// Handle verifies and processes incoming messages according to the protocol.
 	// If there is any message produced in reply, it will be send to reply
 	// channel, otherwise nil channel is returned. The return value new
--- a/client/client.go
+++ b/client/client.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/hyperledger-labs/minbft/api"
 	"github.com/hyperledger-labs/minbft/client/internal/requestbuffer"
+	"github.com/hyperledger-labs/minbft/messages"
 
 	logging "github.com/op/go-logging"
 )
@@ -49,6 +50,15 @@ type Client interface {
 	Request(operation []byte) (resultChan <-chan []byte)
 }
 
+// ReplyCollector processes replies and extracts results
+//
+// Collect gathers Reply messages received from
+// the passed channel, finishes the request processing, and sends the
+// result of request execution to the passed channel.
+type ReplyCollector interface {
+	Collect(in <-chan *messages.Reply, out chan<- []byte, remover requestRemover)
+}
+
 // New creates an instance of Client given a client ID, total number
 // of replica nodes n, number of tolerated faulty replica nodes f, and
 // a stack of external interfaces.
@@ -64,7 +74,34 @@ func New(id uint32, n, f uint32, stack Stack) (Client, error) {
 	}
 
 	seq := makeSequenceGenerator()
-	return makeRequestHandler(id, seq, stack, buf, f), nil
+	collector := makeReplyCollector(f)
+	return makeRequestHandler(id, seq, stack, collector, buf, f), nil
+}
+
+// New creates an instance of Client given a client ID, total number
+// of replica nodes n, number of tolerated faulty replica nodes f, and
+// a stack of external interfaces.
+func GetClient(id uint32, n, f uint32, stack Stack, collector ReplyCollector) (Client, error) {
+	if n < f*2+1 {
+		return nil, fmt.Errorf("Insufficient number of replica nodes")
+	}
+
+	buf := requestbuffer.New()
+
+	if err := startReplicaConnections(id, n, buf, stack); err != nil {
+		return nil, fmt.Errorf("Failed to initiate connections to replicas: %s", err)
+	}
+
+	seq := makeSequenceGenerator()
+	return makeRequestHandler(id, seq, stack, collector, buf, f), nil
+}
+
+// GetMinBFTCollector constructs a replyCollector compatible with MinBFT
+// using the supplied tolerance to remove the request from when its
+// processing is finished.
+func GetMinBFTCollector(f uint32) ReplyCollector {
+
+	return makeReplyCollector(f)
 }
 
 // Request implements Client interface on requestHandler
--- a/client/client.go
+++ b/client/client.go
@@ -23,7 +23,6 @@ import (
 
 	"github.com/hyperledger-labs/minbft/api"
 	"github.com/hyperledger-labs/minbft/client/internal/requestbuffer"
-	"github.com/hyperledger-labs/minbft/messages"
 
 	logging "github.com/op/go-logging"
 )
@@ -50,15 +49,6 @@ type Client interface {
 	Request(operation []byte) (resultChan <-chan []byte)
 }
 
-// ReplyCollector processes replies and extracts results
-//
-// Collect gathers Reply messages received from
-// the passed channel, finishes the request processing, and sends the
-// result of request execution to the passed channel.
-type ReplyCollector interface {
-	Collect(in <-chan *messages.Reply, out chan<- []byte, remover requestRemover)
-}
-
 // New creates an instance of Client given a client ID, total number
 // of replica nodes n, number of tolerated faulty replica nodes f, and
 // a stack of external interfaces.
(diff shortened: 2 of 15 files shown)


client/client.go
```diff
@@ -64,7 +74,34 @@ func New(id uint32, n, f uint32, stack Stack) (Client, error) {
...
 	seq := makeSequenceGenerator()
-	return makeRequestHandler(id, seq, stack, buf, f), nil
+	collector := makeReplyCollector(f)
+	return makeRequestHandler(id, seq, stack, collector, buf, f), nil
+}
+
+// New creates an instance of Client given a client ID, total number
+// of replica nodes n, number of tolerated faulty replica nodes f, and
+// a stack of external interfaces.
+func GetClient(id uint32, n, f uint32, stack Stack, collector ReplyCollector) (Client, error) {
```
nhoriguchi: We already defines `api.ReplicaConnector` in `type Stack`, so I think that we can choose needed replica collector via that interface.

sergefdrv:
@jcs47 First, a few general remarks. Please follow [the convention](https://git-scm.com/docs/git-commit#_discussion) to begin a commit message with a short commit title. (This reminds me of #28...) You probably wanted to submit a [_draft_ pull request](https://help.github.com/en/articles/about-pull-requests#draft-pull-requests).

core/replica.go
```diff
@@ -51,6 +52,33 @@ type Replica struct {
 	handleStream messageStreamHandler
 }
 
+// GetReplica creates a new instance of replica node
+func GetReplica(id uint32, configer api.Configer, stack Stack, handler api.ProtocolHandler, logger *logging.Logger) (*Replica, error) {
```
nhoriguchi: This also seems to me to introduce code duplication, maybe adding `api.ProtocolHandler` into `type replicaStack` is more suitable.

nhoriguchi (commented):
Hi @jcs47, thank you for paying attention to this project. I left a few comments, but I still don't have the whole idea about how this new feature is useful. So if you revise the PR, please consider to separate the change into a series of small patches. Then, you will get more better feedback.

client/client.go
sergefdrv: @Naoya-Horiguchi I guess the core idea here was to make `GetClient` a generic constructor for client part of BFT protocol. Then it be specialized with a specific `ReplyConnector` implementation, according to the specific BFT protocol used.

core/replica.go
sergefdrv: Good point. This function is almost identical to `New`. If this proposal passes, it would make sense to simply call `GetReplica` from `New`.

client/client.go
```diff
@@ -49,6 +50,15 @@ type Client interface {
 	Request(operation []byte) (resultChan <-chan []byte)
 }
 
+// ReplyCollector processes replies and extracts results
+//
+// Collect gathers Reply messages received from
+// the passed channel, finishes the request processing, and sends the
+// result of request execution to the passed channel.
+type ReplyCollector interface {
+	Collect(in <-chan *messages.Reply, out chan<- []byte, remover requestRemover)
```
sergefdrv: Note that this interface depends on MinBFT-specific definition of Reply message, as well as on unexported `requestRemover` type.

client/request.go
```diff
@@ -35,11 +35,12 @@ type requestHandler func(operation []byte) <-chan []byte
 // makeRequestHandler constructs a requestHandler uisng the supplied
 // clientID, sequence number generator, authenticator, request buffer,
 // and number of tolerated faulty replicas.
-func makeRequestHandler(clientID uint32, seq sequenceGenerator, authen api.Authenticator, buf *requestbuffer.T, f uint32) requestHandler {
+func makeRequestHandler(clientID uint32, seq sequenceGenerator, authen api.Authenticator, collector ReplyCollector, buf *requestbuffer.T, f uint32) requestHandler {
 	submitter := makeRequestSubmitter(clientID, seq, authen, buf)
-	collector := makeReplyCollector(f, buf)
```
sergefdrv: Maybe we should rather pass `requestRemover` to `makeReplyCollector`.

client/request.go
```diff
@@ -48,34 +49,44 @@ func makeRequestHandler(clientID uint32, seq sequenceGenerator, authen api.Authe
...
-	remover := makeRequestRemover(buf)
-	return func(in <-chan *messages.Reply, out chan<- []byte) {
-		collectReplies(f, in, remover, out)
+// tolerance to remove the request from when its processing is finished.
+func makeReplyCollector(f uint32) ReplyCollector {
+
+	return &replyCollector{
+
+		faults: f,
+		action: collectReplies,
```
sergefdrv: `collectReplies` could simply be a method attached to `replyCollector` type.

api/api.go
```diff
@@ -22,6 +22,16 @@ import (
 	"time"
 )
 
+// ProtocolHandler handles incoming message according to the implemented protocol.
+//
+// Handle verifies and processes incoming messages according to the protocol.
+// If there is any message produced in reply, it will be send to reply
+// channel, otherwise nil channel is returned. The return value new
+// indicates that the message has not been processed before.
+type ProtocolHandler interface {
```
sergefdrv, line 26: This package was supposed to define the API between MinBFT core and external components, but not components of the core. If we decide to expose those components, it would be better to move those to separate packages, e.g. under `core`. Then those packages would define their API.

core/replica.go
```diff
@@ -51,6 +52,33 @@ type Replica struct {
...
+		return nil, fmt.Errorf("%d nodes is not enough to tolerate %d faulty", n, f)
+	}
+
+	replica := &Replica{
+		id: id,
+		n:  n,
+
+		stack: stack,
+
+		log: messagelog.New(),
```
sergefdrv: This field is used in `replica.start`, but it is not accessible by `handler`.

core/replica.go
```diff
@@ -51,6 +52,33 @@ type Replica struct {
...
+
+		handleStream: makeMessageStreamHandler(handler, logger),
```
sergefdrv: Basically, this would allow to reuse code from `makeMessageStreamHandler`, but that function depends on message definitions which are MinBFT-specific. Perhaps, if we make the message definition generic enough, that function would become quite trivial.

sergefdrv (changes requested):
If I understood correctly, the main idea here was to separate MinBFT-sepecific implementation from more generic code that interacts with the external interfaces. This could potentially allow to reuse those parts separately elsewhere. So far, the "generic" part does not appear really generic; and it is not quite clear yet if that would eventually reduce to something rather trivial. I was also not sure how that would agree with the scope of the project. Nevertheless, such restructuring might potentially improve overall code structure.

sergefdrv:
> I was also not sure how that would agree with the scope of the project.

@ynamiki @Naoya-Horiguchi What do you think?

client/client.go
jcs47: I see two options for Reply message: either change the method signature to  `Collect(in <-chan interface{}, out chan<- []byte, remover requestRemover)` (which would require additional changes throughout the code) or simply adding an extra field to the Reply message (e.g., `metadata []byte`) that would allow this type to be more generic.

Regarding the unexported `requestRemover` type, I thought about changing into to be part of the external API, but since it is created using an object from the `client/internal` package, I thought it would be best to leave it unexported.

client/request.go
jcs47: But that would force us to expose the RequestBuffer declared in `client/internal`, which is an object that is shared by requestRemover and other parts of the code that we do no want to expose. A developer that would want to supply its own ReplyCollector would have to be aware of this.

api/api.go
jcs47: I am not sure I understand the idea. Wouldn't be enough to just move `ProtocolHandler` to `core`?

core/replica.go
jcs47: It was already like this in the master repository.

core/replica.go
jcs47: I think the workflow implemented by the function is both generic and useful enough, maybe it suffices to adapt it to work with `interface {}` instead of `*messages.Message`?

client/request.go
jcs47: Yeah, good point.

client/client.go
jcs47: @Naoya-Horiguchi That is a good point. Maybe it would be better to supply the collector in the stack type.
@sergefdrv , do you mean `ReplyConnector`? if, so, yes, that is the idea.

core/replica.go
jcs47: Yes, adding the ProtocolHandler to replicaStack is a good design choice, but my idea was to preserve the constructor that allowed us to obtain an instance of Replica that would adopt the MinBFT protocol by default, while also providing a more generic constructor that would allow a third party to supply its own protocol. 

The key idea here is to render replica connectivity as decoupled as possible from the ordering protocol, with the goal to make that connectivity re-usable for other projects that might adopt another ordering protocol different from minbft.

client/client.go
sergefdrv: As far as I can see, `GetClient` is supposed to be a generic version of `New`. If we change the definition of `Stack`, we wouldn't have a default constructor for MinBFT client, i.e. `New` would become the same as `GetClient`. We probably want to keep a default constructor for MinBFT.

client/client.go
sergefdrv: > change the method signature to `Collect(in <-chan interface{}, out chan<- []byte, remover requestRemover)`

This seems to be more aligned with the whole idea of making implementation of internal connectivity generic.

> Regarding the unexported requestRemover type ... it would be best to leave it unexported.

I'm afraid external packages wouldn't be able to use the new interface because they have no access to the unexported type.

client/client.go
jcs47: Yes, that was the intention. To preserve a default constructor for MinBFT and another to supply another protocol.

client/request.go
sergefdrv: If we leave it as is, the user of the proposed generic connectivity would have no control over this protocol-specific functionality. A generic code should not force MinBFT-specific `requestRemover`.

api/api.go
sergefdrv: Either to `core` or maybe better to a new sub-package under `core`. That sub-package might encapsulate all related code, then.

client/client.go
jcs47: > I'm afraid external packages wouldn't be able to use the new interface because they have no access to the unexported type.

Do you think it would be wise to make the RequestRemover external?

core/replica.go
sergefdrv: Yes, but generic connectivity should also allow to somehow distribute messages generated by `ProtocolHandler`.

client/request.go
jcs47: Then I wonder if we should make RequestRemover external.

core/replica.go
sergefdrv: I suggest you try and see if it works.

client/request.go
sergefdrv: If you envision how MinBFT-specific request remover may be generically useful.

core/replica.go
sergefdrv: This might even be the most critical piece of feedback to start addressing.

jcs47: The solution I proposed seems to work. Please check commit 40ba273008f3263bd43686648006b1fe82df0fdb, which provides additional methods to the ProtocolHandler interface to serialize/deserialize messages. Don't mind the absence of documentation for the Wrap method, it is dealt with in the subsequent commit.

api/api.go
jcs47: Wouldn't it make sense to put `ProtocolHandler` under `core` and move the code specific to minbft to either `core/minbft` or `core/default`?

client/request.go
jcs47: I'm thinking that maybe the better approach would be to enable developers to supply their own RequestHandler instead of a ReplyCollector. That way usage of a RequestRemover is at the discretion of the programmers, using whatever interface they want. It would also allow to specify how the request message is defined.

jcs47: Upon more exploration and reflection, I think there is no point in trying to make `client.New()` more generic. Developers can simply create their own implementation of the `Client` interface and re-use `ReplicaConnector`.

core/replica.go
sergefdrv: It seems to be a bit too much fuss around trying to reuse such a simple function as `makeMessageStreamHandler`. The external API tries to be generic enough to handle most of generic stuff. Maybe try to push some more generic code behind the external API? Nevertheless, if we are to go with the original proposal, it might make sense to start with some refactoring of the original code. 

Firstly, we could define a high-level abstract message type (interface) that knows how to Marshal/Unmarshal itself, as well as to give its human-readable text representation. Then this would be the job of the message implementation, not of `ProtocolHandler`. To support unmarshalling, `makeMessageStreamHandler` would receive as a parameter a mechanism to create empty messages (message factory). Empty messages would be initialized by unmarshalling the received bytes.

Secondly, we could allow for any number of reply messages, not just only one. Think of reversing the direction of peer message flow. A replica would receive from its peer just a single Hello message, then reply with a stream of messages from its message log.

sergefdrv:
@jcs47 FYI, to maintain a coherent `git blame`-able history, we do not address review comments by adding more follow-up commits. I will soon prepare a proper contribution guideline (#28). The best we have so far is the discussion in #12.

core/replica.go
jcs47: After checking out your pull request, I think an even simpler approach would be to just use a `Replica` interface. So instead of using `minbft.New`, developers that just want to re-use the connectivity could just make their own implementation of that interface. I was trying to make this in such a way that we could explicitly delimit the protocol in an interface, but maybe this is simply too much of a fuss like you speculate.

sergefdrv: So what's your plan regarding this PR?

jcs47: I would suggest your PR to take priority over this one. Once it is merged with the master branch, l will pull the commits and work with what you added. If I find that none of what I proposed in ths PR is useful after the update, then this should simply be closed.

ynamiki:
I would like to focus on the core protocol part; we have many features to be implemented for practical use in the part. And I am afraid that it is too early to generalize based on current implementation, which is under the development. Sorry for my late response.

> > I was also not sure how that would agree with the scope of the project.
> 
> @ynamiki @Naoya-Horiguchi What do you think?

sergefdrv:
> I would like to focus on the core protocol part; we have many features to be implemented for practical use in the part. And I am afraid that it is too early to generalize based on current implementation, which is under the development.

Should we close this PR?

ynamiki:
@Naoya-Horiguchi I would like to hear your comments.

nhoriguchi:
> @Naoya-Horiguchi I would like to hear your comments.

I'm OK to close this PR now because base code is updated (not because the original proposal is rejected). This feature is not in the roadmap, but that's not a good reason to reject it. So we can simply wait that new version of PR is submitted.

sergefdrv:
@jcs47 We close this PR as outdated. Please submit an updated version, if you decide to.

PR #102, opened 2019-07-12
Redesign connectivity

sergefdrv:
The main goal of this pull request is to improve external interface of connectivity.

The first few commits make API documentation more specific and particularly require authentication of connection destination. Then, more interfaces are defined to clarify how replica support handling of incoming connections. After some prompted refactoring, the API is extended to allow replicas distinguish between incoming peer and client connections. Please find more details in corresponding commit messages.

Finally, the last commit (with RFC tag) suggests to require mutual authentication between replicas. This deserves more consideration since it adds more complexity burden onto external component implementation. An alternative approach would be to reverse the direction of communication between replicas. Namely, when a replica initiates a connection to its peers, as opposed to feeding its messages there, it would only receive messages supplied by the connection destinations. Since all replicas are fully connected, it would be roughly equivalent to establishing mutually authenticated connections, but only requires to authenticate destinations, which seems to be less complicated to achieve.


sergefdrv:
Sorry, commits are in showed by GitHub in wrong order...

jcs47:
This PR updates sections of the code that I am not yet too familiar with, but I agree with the functionality that it introduces. But I am not sure I understand how the alternative approach to mutual authentication would work. If each process would only be able to receive data from the destination it connects to, wouldn't it be necessary to establish two separate connections between each pair of processes? In any case, I think the way it is implemented in this PR is the way to go.

sergefdrv:
@jcs47 

> This PR updates with sections of the code that I am not yet too familiar with, but I agree with the functionality that it introduces.

I would be grateful if you could also have a deep look and leave detailed feedback.

>  If each process would only be able to receive data from the destination it connects to, wouldn't it be necessary to establish two separate connections between each pair of processes? 

Yes, each pair of replicas anyway establish cross-connections to each other. If just the connection destination is authenticated then two unidirectional cross-connections virtually provide a mutually-authenticated bidirectional channel. 

The only difference from a truly mutually-authenticated channel I can think of is that two independent connections are not automatically bound to each other. So a message passed in one direction would not automatically share the same communication context with messages passed in the opposite direction. Do you think this difference might have some effect in our case?

> In any case, I think the way it is implemented in this PR is the way to go.

I'm still in doubt. It puts an extra requirement onto external code. Implementing mutual authentication doesn't seem as easy as normal destination-only authentication. This added complexity might be error-prone. If there's no good reason to require truly mutually-authenticated channels, as discussed above, I now tend to rather drop the last commit and adopt the alternative approach.

jcs47:
> I would be grateful if you could also have a deep look and leave detailed feedback.

One thing that I noticed is that you converted the Replica struct into an interface. I am thinking that this might be provide a much simpler way to achieve the generic connectivity that I was trying to do. It would be a matter of developers creating their own implementation of this object.

> The only difference from a truly mutually-authenticated channel I can think of is that two independent connections are not automatically bound to each other. So a message passed in one direction would not automatically share the same communication context with messages passed in the opposite direction. Do you think this difference might have some effect in our case?

None that I can think off. I just wonder if this approach does not create too many file descriptors, specially with dozens of peers. Is this approach supposed to be more efficient at the cost of more resources?

> I'm still in doubt. It puts an extra requirement onto external code. Implementing mutual authentication doesn't seem as easy as normal destination-only authentication. This added complexity might be error-prone. If there's no good reason to require truly mutually-authenticated channels, as discussed above, I now tend to rather drop the last commit and adopt the alternative approach.

I don't think this is a huge downside, but if the alternate approach does not fundamentally change how the connectivity is used, maybe it worth going for it.

sergefdrv:
@jcs47

> I just wonder if this approach does not create too many file descriptors, specially with dozens of peers. Is this approach supposed to be more efficient at the cost of more resources?

The approach (of unidirectional authenticated cross-channels) assumes each replica would still connect to each other only once (no additional connection when handling incoming ones). Only the direction of message flow would change so that the peer messages come from an authenticated connection target rather than from an unauthenticated connection source.

api/api.go
```diff
@@ -45,29 +50,39 @@ type Configer interface {
 
 //======= Interface for module 'network' =======
 
+// ConnectionHandler handles incoming connections.
+//
+// PeerMessageStreamHandler method provides a mechanism to initiate
+// message exchange with the specified peer replica. It must only be
```
nhoriguchi, line 56: Replica is a core concept of BFT algorithm, but sometimes seem to be called like "peer" or "peer replica". Do we call them to represent exactly the same things, or use them differently for the situations?

sample/conn/common/connector/connector.go
```diff
@@ -0,0 +1,90 @@
...
+// can be used.
+//
+// AssignReplica method assigns a replica representation to the
+// replica identifier.
+type ReplicaConnector interface {
+	api.ReplicaConnector
+	AssignReplica(id uint32, replica api.ConnectionHandler)
+}
+
+type common struct {
```
nhoriguchi, line 34: The name `common` sounds a little too generic at a glance, although I see from the context that it means "common connector".

nhoriguchi (commented):
As for [RFC] part, I think that some authentication mechanism in minbft might be useful, but this project focuses on a consensus component and it's assumed to be used by more generic products (like blockchain platforms or other distributed computing platforms) which should have more robust authentications on their own. So it's close to "out of scope" at this point.

Nonetheless blacklisting Byzantine faulty replica sounds interesting to me, because it's lightweight (maybe just changing around `peerState`) and could be addressed as optimization.

Anyway, I looked through the change set and it looks good me (comments are marginal). Thank you for the work.

sergefdrv:
> As for [RFC] part, I think that some authentication mechanism in minbft might be useful, but this project focuses on a consensus component and it's assumed to be used by more generic products (like blockchain platforms or other distributed computing platforms) which should have more robust authentications on their own. So it's close to "out of scope" at this point.

TL;DR: I plan to drop the RFC commit.

The actual motivation for this was to better express requirements on the external components. Therefore, the absolute requirement to authenticate the conneciton destination was clearly documented. 

It seems also desirable to distinguish client/peer connections at the core level. Consider, we want to reverse the direction of message flow between replicas. (This has its benefits.) Then the ability to distinguish client/peer connections would prevent clients from pretending as a peer replica and receiving messages they are not supposed to see, like requests of other clients embedded in Prepare messages. It would then be up to the external code to find a proper way of distinguishing client/peer connections, depending on the use-case requirements. For example, it could use mutual TLS, or separate networks, or maybe even no enforcement.

The RFC part considers requiring the external code to provide precise mutual authentication between replicas. Now, I think it would, indeed, be an excessive requirement. Mutual authentication seems to be the only choice if we cannot establish a connection back, like in a client-server architecture where clients might not have public network addresses to connect to. But in our case, replica are always interconnected in both directions. So one-side authentication seems to be enough.

api/api.go
sergefdrv: Although, I'm not sure if that is a correct way, I tend to use "replica" when referring to _any_ replica, and "peer replica" (or shortly "peer") when referring to a replica _other_ than the current one. In different words, "replica" is used from the external or client's perspective, whereas "peer" from a replica's point of view.

@jcs47 What do you think?

sample/conn/common/connector/connector.go
sergefdrv: It is supposed to represent common pieces of client- and replica-side implementation.

sample/conn/grpc/server/server.go
```diff
@@ -99,7 +117,7 @@ func (s *server) Chat(stream proto.Channel_ChatServer) error {
 			if err == io.EOF {
 				break
 			} else if err != nil {
-				err = fmt.Errorf("error receiving from replica server stream: %s", err)
+				err = fmt.Errorf("Error receiving from server stream: %s", err)
```
ynamiki, line 136: Note: It is said that error strings should not be capitalized, but our project use the capitalized style consistently.
https://github.com/golang/go/wiki/CodeReviewComments#error-strings

ynamiki (approved):
I have no objection to drop the RFC commit. I think it is good to focus on the core consensus protocol at least for now, as Naoya mentioned. I left a comment, but it is not a request for change. Thank you for your work.

sample/conn/grpc/server/server.go
sergefdrv: Thanks for pointing that out, I didn't know that convention. I think it would be better to keep a consistent style. Maybe we can change the style to recommended as a separate low-priority issue.

sergefdrv: What we could also do is to use "github.com/pkg/errors" package for wrapping error messages in the context of the call stack.

PR #113, opened 2019-08-07
Fix a misspelling in a variable name

ynamiki:
There is a misspelling in a variable name of a header file.
This PR fixes that.


nhoriguchi (approved):

sergefdrv (approved):
Thanks for spotting this!

PR #114, opened 2019-08-08
Fix Memory Leak

ynamiki:
C.CString(), which convert a Go string to a C string, allocates the C string in the C heap. We have to free it.

https://github.com/golang/go/wiki/cgo#go-strings-and-c-strings

> One important thing to remember is that C.CString() will allocate a new string of the appropriate length, and return it. That means the C string is not going to be garbage collected and it is up to you to free it. A standard way to do this follows.


ynamiki:
I confirmed that the code is the only use of C.CString() in our project; no other use of the function.

nhoriguchi (approved):
Looks good to me, thank you.

sergefdrv (approved):
Thanks!

Issue #87, opened 2019-04-24
Make usig/sgx package retrievable from the repository


sergefdrv:
Now the main code base is converted to a Go module. However, we cannot fully utilize flexibility and convenience of Go module system due to the fact that `usig/sgx` is still a special package. That is because `usig/sgx/usig-enclave.go` has to be generated with `make`. This means a module user cannot retrieve this package directly from the repository, but needs to build it separately. We should change the way we build the enclave to make this an ordinary Go package.

To achieve this, we need to break the dependency of cgo code on SGX SDK dynamic libraries. In other words, building of the encalve-related C stuff needs to be split from the main Go code base.

Making the enclave wrapper (shim) library a dynamically-linked module (using `dlopen()` in cgo code) seems to be a good option to achieve this.

For more details, please refer to [the original discussion starting here](https://github.com/hyperledger-labs/minbft/pull/72#issuecomment-481881753).

nhoriguchi:
There's a ongoing discussion about Go-plugin, but I think this specific issue is solved by #115. So let's close this.

PR #115, opened 2019-08-08
Make this module able to build as an ordinal Go module

ynamiki:
As described in #87, generating a Go file in the Makefile is a barrier to handle this module as an ordinal Go module. In this patch, we change the structure not to require such process.

The cause of the problem is that we need to build and link libraries containing "untrusted" part of codes for Intel SGX for both simulation mode and hardware mode separately. Now we change to dynamically link the library so that we can build the Go part of codes without the code generation at Make time.


ynamiki:
@sergefdrv @Naoya-Horiguchi This PR is not finished, but I would like to ask you for your review whether it is on a reasonable way.

ynamiki:
The test fails because we need to add the path for libusig_shim.so to `$LD_LIBRARY_PATH`. It does not have an impact for the code design.

usig/sgx/usig-enclave.go
```diff
@@ -16,12 +16,69 @@
...
 package sgx
 
-// #cgo pkg-config: $SGX_URTS_PACKAGE
-// #cgo CFLAGS: -I${SRCDIR}/shim
-// #cgo LDFLAGS: -L${SRCDIR}/shim -lusig_shim
-//
-// #include <stdlib.h>
-// #include "usig.h"
+/*
+// TODO Use pkg-config?
```
nhoriguchi: I'm not sure that pkg-config helps here, but if you choose to drop it, you can also remove lines about pkg-config in `README.md` and `.circleci/config.yml`.

nhoriguchi (commented):
I confirmed that package delivery via repository worked well, and if I read correctly the proposed changes straightforwardly do what we need. I left one comment, but mainly I agree with the current direction.

usig/sgx/usig-enclave.go
```diff
@@ -16,12 +16,69 @@
...
+#cgo CPPFLAGS: -I/opt/intel/sgxsdk/include
```
sergefdrv: Obviously, hard-coding an absolute path is a bit hacky. This line seems to indicate that the cgo code still depends on SGX SDK...

usig/sgx/usig-enclave.go
sergefdrv: To avoid confusion, `pkg-config` would need the exact library name: `libsgx_urts_sim` or `libsgx_urts`, but we just want to get rid of that...

usig/sgx/usig-enclave.go
```diff
@@ -16,12 +16,69 @@
...
+
+// wrappers to call a function through a function pointer, which is returned by
+// dlsym().
+
+sgx_status_t usig_init(const void *pointer,
+                       const char *enclave_file, sgx_enclave_id_t *enclave_id,
+                       void *sealed_data, size_t sealed_data_size)
+{
+	return ((sgx_status_t (*)(const char *, sgx_enclave_id_t *, void *, size_t)) pointer)
+			(enclave_file, enclave_id, sealed_data, sealed_data_size);
```
sergefdrv: It might be a bit more convenient to define a function type:

```c
typedef usig_init_fn(const void *pointer,
                     const char *enclave_file, sgx_enclave_id_t *enclave_id,
                     void *sealed_data, size_t sealed_data_size)
```

So the result of `dlsym` can be assigned to a variable of that type. Then the variable could be invoked as normal function.

usig/sgx/usig-enclave.go
```diff
@@ -249,3 +335,49 @@ func sgxUint8SliceToBigInt(sgxArr []C.uint8_t) *big.Int {
 
 	return new(big.Int).SetBytes(bytes)
 }
+
+func dlopen() (unsafe.Pointer, error) {
+	f := C.CString(untrustedFile)
+	defer C.free(unsafe.Pointer(f))
+
+	handle := C.dlopen(f, C.RTLD_LAZY)
```
sergefdrv: Maybe we should use `RTLD_NOW` as in https://golang.org/src/plugin/plugin_dlopen.go.

sergefdrv (commented):
Thanks, that is a great piece of work! I think the general direction is what we want. However, using `dlopen` directly turns out a bit cumbersome. Maybe we should also consider Go plugin system? In any way, it seems like we'd better make cgo code fully independent from SGX SDK.

usig/sgx/usig-enclave.go
ynamiki: I was considering how we add the path of include files of the SDK without hard-coding. Because now we need only the include path (`-I`) (e.g. no `-L` and `-l`), we don't need the code generation like before (i.e. `#cgo pkg-config: libsgx_urts_sim` works for both modes).

Anyway, I will try to remove dependency to the header files in next revision.

ynamiki:
Thanks for your comments.

I tried [Go plugin](https://golang.org/pkg/plugin/), but found it cannot be used in  our project; a plugin (`*.so`) must be written in Go (some "[module data](https://github.com/golang/go/blob/4ce6a8e89668b87dce67e2f55802903d6eb9110a/src/runtime/plugin.go#L20)" is required in a  plugin).

I will refine my patch.

usig/sgx/usig-enclave.go
ynamiki: I will update with `RTLD_NOW|RTLD_GLOBAL`, same as `plugin`.

`LAZY` has better performance, but `NOW` is safer in terms of getting an error at `dlopen()` for undefined symbols. If there are many functions which is not referenced, using `NOW` has negative impact on performance. But I think this is not applicable to this project.

sergefdrv:
> I tried [Go plugin](https://golang.org/pkg/plugin/), but found it cannot be used in our project; a plugin (`*.so`) must be written in Go (some "[module data](https://github.com/golang/go/blob/4ce6a8e89668b87dce67e2f55802903d6eb9110a/src/runtime/plugin.go#L20)" is required in a plugin).

The suggestion was actually to consider writing a Go plugin in Go. Though, I am not sure which approach would eventually better fit in our case.

usig/sgx/usig-enclave.go
sergefdrv: Removing that dependency would require to avoid using SGX SDK types in `usig.h`. That probably also means translating parts of Go code from `usig-enclave.go` to C code in `shim.c`.

ynamiki:
Reopen; I have accidentally closed this issue.

ynamiki:
@sergefdrv @Naoya-Horiguchi I have updated the code. Could you review the latest version?

Now we can build the Go part without the special build process that includes code generation. A user will be able to import this module same as a general Go module.

Note that, this PR removes the special build process, but still we have dependency on the header files of SGX SDK. To remove the dependency we need to change the interface between Go and C (the unstusted part) to get rid of data types such as `sgx_status_t` provided by SGX SDK. This is future work.

The followings are changes from previous revision:
- Update Makefile to install the library which is newly introduced by separating from the main code.
- Update README to update `LD_LIBRARY_PATH` to include a path for the library.
- Use `__typeof__` (GNU extension) to get type of functions defined in the library.

README.md
```diff
@@ -189,6 +189,14 @@ argument. Sample configuration files can be found in
 `sample/authentication/keytool/` and `sample/peer/` directories
 respectively.
 
+Before running the example, the environment variable `$LD_LIBRARY_PATH`
+needs to be updated to include a directory of `libusig_shim.so`,
+which is installed to the following directory by `make install`.
```
nhoriguchi: Precisely saying, the following code block is to update the environment variable, not showing a directory (so referring to by the expression "the following directory" sounds unfit to me). So how about this?

> Before running the example, the environment variable `$LD_LIBRARY_PATH` needs to include `sample/lib` where `libusig_shim.so` is installed by `make install`.

... by the way, `sample/lib` is a correct path for libusig_shim.so? It seems that it's located under `$PWD/usig/sgx/shim` as your change on `.circleci/config.yaml` sugguests.

usig/sgx/usig-enclave.go
```diff
@@ -16,12 +16,66 @@
...
 package sgx
 
-// #cgo pkg-config: $SGX_URTS_PACKAGE
-// #cgo CFLAGS: -I${SRCDIR}/shim
-// #cgo LDFLAGS: -L${SRCDIR}/shim -lusig_shim
-//
-// #include <stdlib.h>
-// #include "usig.h"
+/*
+// We use only a value of `Cflags` from pkg-config.
```
nhoriguchi, line 20: "Cflags" looks weird to me , you mean "CFLAGS"?

nhoriguchi (commented):
@ynamiki, thank you for your effort.

> A user will be able to import this module same as a general Go module.

Really nice, and I agree with separating the header dependency issue into a separate one.

There's a typo in the description of [3253ccc7](https://github.com/hyperledger-labs/minbft/pull/115/commits/3253ccc78b255a7513bd2d2864852ea32da5901b) ("usig/sgx/usig-enclable.go"). And I left some comment/description issues.

usig/sgx/usig-enclave.go
ynamiki: It comes from the notation of [pkg-config](http://manpages.ubuntu.com/manpages/bionic/man1/pkg-config.1.html).

By the directive `#cgo pkg-config: libsgx_urts_sim`, Cgo read values from `/opt/intel/sgxsdk/pkgconfig/libsgx_urts_sim.pc`, which has the following lines:

```
prefix=/opt/intel/sgxsdk
includedir=${prefix}/include
libdir=${prefix}/lib64

Name: libsgx_urts_sim
Description: SGX uRTS simulation library
Version: 2.3.101.46683
Cflags: -I${includedir}
Libs: -L${libdir} -lsgx_urts_sim
Libs.private: -lsgx_uae_service_sim
```

There are options to be given to a compiler and a linker to use this library. In our case, we need only the value of `Cflags` because we don't link the library at the built time. And the values is the same as in `/opt/intel/sgxsdk/pkgconfig/libsgx_urts.pc`, which is for the hardware mode.

nhoriguchi: OK, thank you for the detail. I have no objection.

README.md
ynamiki: Right, I will update the description in the next revision.

libusig_shim.so is placed in `usig/sgx/shim` by `make build`, and is copied to `sample/lib` by `make install` (`peer` and `keytool` are also copied from a build directory). The install task is aimed for collecting binaries in a single place (i.e. `sample`) and it is not executed in the CI process. So the CI refers `usig/sgx/shim` and the README refers `sample/lib`.

usig/sgx/usig-enclave.go
```diff
@@ -16,12 +16,66 @@
...
+// So, this works also for the hardware mode.
+#cgo pkg-config: libsgx_urts_sim
```
sergefdrv: Conversely, why don't use `libsgx_urts`?

sergefdrv (commented):
Sorry for the delay. It is certainly a big improvement in the build process. I suggest to address @Naoya-Horiguchi's comment on the read-me update, then we should be ready to merge it.

usig/sgx/usig-enclave.go
ynamiki: I don't mind `_sim` or not. I will change to more fundamental one `libsgx_urts`.

ynamiki:
@sergefdrv @Naoya-Horiguchi Thanks for your comments. I have update the code; the followings are changes from the previous revision.

- Fix typo in a commit message: https://github.com/hyperledger-labs/minbft/pull/115#pullrequestreview-285337185
- Refine README: https://github.com/hyperledger-labs/minbft/pull/115#discussion_r322109639
- Update `#cgo pkg-config` directive to refer `libsgx_urt`: https://github.com/hyperledger-labs/minbft/pull/115#discussion_r326173317

nhoriguchi (approved):
Looks good to me, thank you.

sergefdrv (approved):
Thanks!

PR #117, opened 2019-08-14
core/utils: Introduce shortString()

### LINKED ISSUE #111: Simplify messageString() output for large message payload
I noticed in "MinBFT on Fabric" trial that orderer's log output was filled with message payload, which was a bit cumbersome to handle. Showing raw payload might be good for some simple workload, but some digest might more preferable for messages with large payload. One possible way to go is to introduce a threshold and to show a digest value if the size of message payload goes above it.

nhoriguchi:
This pull request suggests to introduce an upper limit of payload string in replica's log.

Resolves #111

### DIFF
--- a/core/utils.go
+++ b/core/utils.go
@@ -75,6 +75,13 @@ func isPrimary(view uint64, id uint32, n uint32) bool {
 	return uint64(id) == view%uint64(n)
 }
 
+func shortString(msg string, max int) string {
+	if len(msg) > max {
+		return msg[0:max-1] + "..."
+	}
+	return msg
+}
+
 func messageString(msg interface{}) string {
 	var cv uint64
 	if msg, ok := msg.(messages.MessageWithUI); ok {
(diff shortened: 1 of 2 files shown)


core/utils.go
```diff
@@ -75,6 +75,13 @@ func isPrimary(view uint64, id uint32, n uint32) bool {
 	return uint64(id) == view%uint64(n)
 }
 
+func getTrimmedMsg(msg string) string {
```
ynamiki: How about naming this function just "trim" or "trimMessage"? I imagine retrieving a value from a member variable of a object from the word "get".

nhoriguchi: Thanks, I'll do it.

sergefdrv: One more option for naming: `shortString`.

core/utils.go
```diff
@@ -75,6 +75,13 @@ func isPrimary(view uint64, id uint32, n uint32) bool {
...
+	if len(msg) > logMaxPayloadWidth {
```
sergefdrv: There is a bit of dissonance between the function name and the name of this constant. The constant name relates to the payload, whereas the function name doesn't seem to reflect that. Maybe rename the constant more generically, or pass its value as a parameter.

nhoriguchi:
Thank you for comments, I've updated the branch by force-pushing the branch. The summary of updates is:

- rebased onto the latest master branch
- renamed the function and the constant
- introduced another parameter `max` to `shortString()`.

The diff between v1 and v2 is [here](https://github.com/Naoya-Horiguchi/minbft/compare/3971cbb...c483293), which might not be easy to read due to the rebase.

nhoriguchi:
I force-pushed again to fix DCO failure. No functional change.

sergefdrv (approved):

ynamiki (approved):

Issue #110, opened 2019-07-30
Fix viewstate package


sergefdrv:
Sometimes, the package test fails: https://circleci.com/gh/hyperledger-labs/minbft/159. This is either a bug in the implementation or in the test that needs to be fixed.

PR #122, opened 2019-10-15
Message interfaces

sergefdrv:
This pull request introduces an abstract interface between a particular representation of protocol messages and their usage by the core protocol logic. 

This decouples the core protocol code from a particular way the protocol messages are represented, e.g. from automatically generated code of Protocol Buffers. It is advantageous to make this separation before introducing more complex message types as those of the view change operation.


core/internal/messagelog/mocks/mock.go
```diff
@@ -34,23 +34,27 @@ func (m *MockMessageLog) EXPECT() *MockMessageLogMockRecorder {
 }
 
 // Append mocks base method
-func (m *MockMessageLog) Append(arg0 *messages.Message) {
+func (m *MockMessageLog) Append(arg0 *protobuf.Message) {
+	m.ctrl.T.Helper()
```
nhoriguchi, line 38: Is this `.Helper()` line relevant to current change? I know this and other changes on mock files are automatically generated by `make generate`, but if so why not included all changes generated by `make generate`? If you have specific reason, please leave some comment on it. Or committing all `make generate` changes is OK, you can do it first at separate patch, then update message interface.

messages/api.go
```diff
@@ -0,0 +1,170 @@
...
+// 	package api
+//
+//	type Sealed interface {
+// 		DoSomething()
+// 		isSealed()
+// 	}
+//
+//	type IsSealed struct{}
+//
+//	func (IsSealed) isSealed() {}
```
nhoriguchi: "Sealed interface" concept is new to me, but it sounds to strictly control visibility of the type, which should be good for code robustness. Anyway this patch only provides interfaces and I can't say anything about how each interface has each method for now.

messages/api.go
```diff
@@ -0,0 +1,170 @@
...
+	NewFromBinary(data []byte) (Message, error)
+	NewRequest(clientID uint32, sequence uint64, operation []byte) Request
+	NewPrepare(replicaID uint32, view uint64, request Request) Prepare
+	NewCommit(replicaID uint32, prepare Prepare) Commit
+	NewReply(replicaID, clientID uint32, sequence uint64, result []byte) Reply
+}
+
+type Message interface {
+	encoding.BinaryMarshaler
+}
```
nhoriguchi, line 33: This works for decoupling logic and representation, so I'm looking forward to see the concrete type.

ynamiki (approved):
Looks good to me.

core/internal/messagelog/mocks/mock.go
sergefdrv: I think I just never looked at changes to auto-generated files. But this commit touches too many files, so the changes should be better minimal. I will update the old generated mock files in a separate commit.

messages/api.go
sergefdrv: This concept is admittedly tricky, but seems to be the most reliable way to enforce explicit implementation of interfaces in Go (see https://github.com/golang/go/issues/34996).

sergefdrv: I was thinking _maybe_ combine those `type IsFoo struct{}` and their methods at the end of the file.

> Anyway this patch only provides interfaces and I can't say anything about how each interface has each method for now.

True, I just wanted to share the interfaces before I proceed with implementation. Now I have some implementation (not yet used), so I can share that, too.

sergefdrv:
Summary of changes in https://github.com/hyperledger-labs/minbft/pull/122#commits-pushed-8900a07:
- Regenerated old mock files in separate commit
- Added message implementation

sergefdrv:
Sorry, I'll fix the DCO issue with the next update...

messages/api.go
nhoriguchi: > This concept is admittedly tricky, but seems to be the most reliable way to enforce explicit implementation of interfaces in Go (see [golang/go#34996](https://github.com/golang/go/issues/34996)).

You're trying to improve Golang itself, wow 😲

nhoriguchi:
> * Added message implementation

Thanks for the update. `Reply` message still unimplemented, is that intentional?

sergefdrv:
> `Reply` message still unimplemented

I seem to have finished implementing messages half-way :slightly_smiling_face: My plan was to get your feedback on this while working on some other patches for view change (that will also fix #110). Those changes might affect how the message interfaces will be used in the core protocol logic.

messages/api.go
sergefdrv: To have some fun at night :laughing:

messages/protobuf/impl.go
```diff
@@ -34,14 +34,20 @@ func (*impl) NewFromBinary(data []byte) (messages.Message, error) {
 		return nil, fmt.Errorf("failed to unmarshal message wrapper: %s", err)
 	}
 
-	switch msg.Type.(type) {
+	switch t := msg.Type.(type) {
+	case *Message_Request:
```
nhoriguchi, line 38: This function switches based on message type extracted from raw message data, but it's seems that the sealed interface `message.IsRequest` is not used here.

https://github.com/sergefdrv/minbft/blob/8245dae27eba1a6bdb4592aac367daec48ba92b8/messages/protobuf/request.go#L22-L25

`protobuf.request` embeds `messages.IsRequest`, so I guess that you realize the new thing on this type. So if you explain how it's used, the concept will get clearer.

messages/protobuf/request_test.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"math/rand"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/hyperledger-labs/minbft/messages"
+)
+
+func TestRequest(t *testing.T) {
```
nhoriguchi, line 26: I guess that some test code using `isRequest()` to check the type elegantly will be added later?

messages/protobuf/impl.go
sergefdrv: That's what I try to convey in [the "explicit interfaces" proposal](https://github.com/golang/go/issues/34996): The sealed interfaces is a tricky solution to the problem.

Here, `msg` is a `Message` structure which is defined by auto-generated code. Protobuf's code generator `protoc` defines a `oneof` field `Message.Type` as an unexported interface `isMessage_Type` with a single unexported method `isMessage_Type()`. The whole purpose of `isMessage_Type` interface is to choose which concrete types can be assigned to the field. This is close to the sealed interfaces construction, but within the same Go package.

Thus, this type switch determines the unmarshaled message type relying on protobuf's definitions. This should work as expected, because the switch cases match concrete types (structures, not interfaces) and `protoc` guarantees the expected semantic of such type switches/assertions. 

However, the returned value of this function is an abstract interface `messages.Message`, defined in `messages/api.go`. That value's type should embed one of `IsXxx` structures defined in `messages/api.go`. The value returned by `newRequest()` in the following line is of `request` type, which embeds `messages.IsRequest`. This embedding ensures that `request` implements the unexported methods of `messages.Request` (the only purpose of `messages.IsRequest`). The remaining methods of `messages.Request` are implemented by `request` itself so that it fulfils `messages.Request` interface.

messages/protobuf/request_test.go
sergefdrv: I think there is no need to test `isRequest()` because `impl.NewRequest()` simply cannot return anything not implementing `messages.Request` interface, which requires `isRequest()` method. It is checked by the compiler. Moreover, `isRequest()` is unexported in `messages` package, so it cannot be invoked here.

sergefdrv:
Summary of [changes](https://github.com/hyperledger-labs/minbft/compare/8245dae27eba1a6bdb4592aac367daec48ba92b8..762ae1a85d75c28f87aba5c8b45b32d9088993f1):

- Fixed DCO issue
- Added implementation of `Reply` message

messages/protobuf/request_test.go
nhoriguchi: OK, so this method is only used to make some type to fulfill the interface. How useful it is will be 
clearer with view-change patch.

nhoriguchi (approved):
The change set looks good to me, so I think it's ready to merge.

> Summary of [changes](https://github.com/hyperledger-labs/minbft/compare/8245dae27eba1a6bdb4592aac367daec48ba92b8..762ae1a85d75c28f87aba5c8b45b32d9088993f1):
> 

This makes me review easier (nice practice), thank you.

sergefdrv:
> The change set looks good to me, so I think it's ready to merge.

I initially planned to merge it after final changes for switching the core to use the message interfaces. That's why I opened this pull request ad "draft".

> This makes me review easier (nice practice), thank you.

In fact, the link is just a copy of the one attached to the "force-pushed" text (which doesn't look like a link):
image

nhoriguchi:
> > The change set looks good to me, so I think it's ready to merge.
> 
> I initially planned to merge it after final changes for switching the core to use the message interfaces. That's why I opened this pull request ad "draft".

"Draft" status was removed when I clicked "ready for review" button to update my review status. I'm not sure that this was the right review process, so sorry if it surprised you. I leave this PR open for you to add the final changes, or to merge current PR branch.

sergefdrv:
It looks like [the official recommendation](https://golang.org/doc/faq#guarantee_satisfies_interface) for resolving ambiguities among similar interfaces, as opposed to using sealed interfaces, is to augment interfaces with so-called qualifying methods. This would look like follows.

```go
type ReplicaMessage interface {
	Message
	ReplicaID() uint32

	ImplementsReplicaMessage()
}

// ...

type Reply interface {
	ReplicaMessage
	// ...

	ImplementsReply()
}
```

```go
type reply struct {
	// ...
}

func (m *reply) MarshalBinary() ([]byte, error) { /* ... */ }

func (m *reply) ReplicaID() uint32 { /* ... */ }

// ...

func (reply) ImplementsReply() {}
func (reply) ImplementsReplicaMessage() {}
```

What do you think of this alternative?

ynamiki:
Thanks for the info. I'd rather like the official way even both are not perfect. It needs less codes, is easier to read for me, and is in the official FAQ.

nhoriguchi:
`Implements*` methods in your example are global so visible to external packages. It seems not what you originally wanted, but if you think that's acceptable, that's fine and I'll go with it.

sergefdrv:
> I'd rather like the official way even both are not perfect. It needs less codes, is easier to read for me, and is in the official FAQ.

I agree that following an official recommendation should make that easier for others to understand it. Though, I'm not sure if it needs less code: Note that `reply struct` needs to implement both `ImplementsReply` and `ImplementsReplicaMessage`. Maybe in our case it's not a big overhead.

> `Implements*` methods in your example are global so visible to external packages. It seems not what you originally wanted, but if you think that's acceptable, that's fine and I'll go with it.

What was desired was to avoid `reply struct` satisfying `messages.ReplicaMessage` interface just because it happens to implement `ReplicaID` method. The "qualifying methods" approach also achieves that goal. The "sealed interfaces" approach is a bit more strict in the sense that _the only way_ to implement `messages.Reply` is to embed `messages.IsReply` provided by `messages` package. So it resembles [nominative typing](https://en.wikipedia.org/wiki/Nominal_type_system) rather than regular Go's [structural typing](https://en.wikipedia.org/wiki/Structural_type_system).

Basically the actual problem I tried to solve is perfectly captured on Wikipedia:

> A pitfall of structural typing versus nominative typing is that two separately defined types intended for different purposes, but accidentally holding the same properties (...), could be considered the same type by the type system, simply because they happen to have identical structure. 

So far, I don't have a strong opinion on which approach to follow. Maybe I adjust the code to follow the "qualifying methods" approach, and we'll see how it looks like. Sleep on it, then decide :slightly_smiling_face:

ynamiki:
> I'm not sure if it needs less code

It is my understanding that the sealed interface needs an additional type definition (`IsReplicaMessage`, etc.), which is not in the official recommended way.

I don't have a strong option and either is fine.

sergefdrv:
I tried to adjust the code for qualifying methods approach. Please have a look to see which approach you prefer.

ynamiki:
Thanks, I prefer the qualifying methods approach.

nhoriguchi:
> Basically the actual problem I tried to solve is perfectly captured on Wikipedia:
> 
> > A pitfall of structural typing versus nominative typing is that two separately defined types intended for different purposes, but accidentally holding the same properties (...), could be considered the same type by the type system, simply because they happen to have identical structure.
> 
> So far, I don't have a strong opinion on which approach to follow. Maybe I adjust the code to follow the "qualifying methods" approach, and we'll see how it looks like. Sleep on it, then decide

OK, so let's go with the "qualifying methods" approach. It's nice to keep the commit about "sealed interface" because the project history has the record and that allows us to revisit easier in the future.

I'm OK to merge your branch, but the latest commit has "RFC" in the subject, so you may make the final change.

sergefdrv:
@ynamiki @Naoya-Horiguchi Thanks for the feedback! The recommended approach indeed doesn't seem too bad, let's stick to it.

> I'm OK to merge your branch, but the latest commit has "RFC" in the subject, so you may make the final change.

I would still prefer to merge it when I make the interfaces actually used by the core... Please let me know if you need it sooner.

sergefdrv:
Summary of [changes](https://github.com/hyperledger-labs/minbft/compare/feb001e808b9bd48ac14f8d5ab939b110d90c334..e94a0665b93329f2dd74e73b519a1b39ccb7a3c3):

- Rebased
- Updated commit message of "messages: Use qualifying methods instead of sealed interfaces"

(More commits will come soon.)

core/message-handling.go
```diff
@@ -417,32 +417,36 @@ func makeUIMessageProcessor(captureUI uiCapturer, processViewMessage viewMessage
...
-			if err := applyReplicaMessage(msg); err != nil {
-				return false, fmt.Errorf("Failed to apply message: %s", err)
+		case *messages.Prepare, *messages.Commit:
+			var view uint64
+
+			switch msg := msg.(type) {
+			case *messages.Prepare:
+				view = msg.View()
+			case *messages.Commit:
+				view = msg.View()
```
ynamiki, line 428: (Solved) Why do we separate these cases? Both have same statement. -> I found that they are updated and have different statements in a next commit.

nhoriguchi:
A few comments:
- I think the final commit ("client,core: Use message interfaces") is the most interesting one since last review, and it looks good to me as long as I checked (it's a little too large to check line-by-line.)
- One question: there are many conversions like `*messages.Request` -> `messages.Request` especially in definitions of function parameters, is that necessary change for your current purpose? Maybe my point is unclear, but I'm a little worried that that could introduce unexpected behaviors.
- Patch 12/13 (["core: Restructure TestMakeReplicaMessageProcessor "](https://github.com/hyperledger-labs/minbft/pull/122/commits/41b32d422bd9d2b99a6327b6fade4643226b2cec)) has some broken indentations, although that will be fixed by patch 13/13.

sergefdrv:
@Naoya-Horiguchi Thanks for reviewing this.

> * I think the final commit ("client,core: Use message interfaces") is the most interesting one since last review, and it looks good to me as long as I checked (it's a little too large to check line-by-line.)

Sorry for making the commit so large, but I couldn't find a better way to do the switch without breaking building/testing.

> * One question: there are many conversions like `*messages.Request` -> `messages.Request` especially in definitions of function parameters, is that necessary change for your current purpose? Maybe my point is unclear, but I'm a little worried that that could introduce unexpected behaviors.

Before the switch, `Request` was a structure (from `messages/protobuf` package), after the switch it is an interface (from `messages` package), thus the declarations change. This must be a pretty safe change, because the compiler detects this kind of type mismatch.

> * Patch 12/13 (["core: Restructure TestMakeReplicaMessageProcessor "](https://github.com/hyperledger-labs/minbft/pull/122/commits/41b32d422bd9d2b99a6327b6fade4643226b2cec)) has some broken indentations, although that will be fixed by patch 13/13.

Thanks, I think I missed it when resolving rebase conflicts. That's easy to fix :slightly_smiling_face:

nhoriguchi:
> Before the switch, `Request` was a structure (from `messages/protobuf` package), after the switch it is an interface (from `messages` package), thus the declarations change. This must be a pretty safe change, because the compiler detects this kind of type mismatch.

Make sense, thank you very much. So I'm totally fine about proposed patches ~~with the indent fix~~(you already did that). :+1:

ynamiki (approved):
I had a question but it is solved.

One (slightly off-topic) question: do we have a nice tool to support refactoring that has been done in the last commit? I guess changing a struct to an interface is a common refactoring method and this cause tedious and extensive code modification. It is hard to do that manually.

nhoriguchi:
> One (slightly off-topic) question: do we have a nice tool to support refactoring that has been done in the last commit? I guess changing a struct to an interface is a common refactoring method and this cause tedious and extensive code modification. It is hard to do that manually.

I have no idea about specific tools, but maybe some modern editor (VSCode in my mind) helps for the purpose. For example, if you change some code around struct definition, VSCode automatically (in real time manner) checks whole repository and marks all type mismatches as warnings. So you can adjust all affected places with confidence.

nhoriguchi:
I tried to resolve conflict, but I found that I had to change `core/request_test.go` (not listed in the conflicting files) to pass the unit tests, so I think we can't simply use [Web UI](https://github.com/hyperledger-labs/minbft/pull/122/conflicts) and we might have to push origin/master via CLI. Should I do this? The [resolved branch](https://github.com/Naoya-Horiguchi/minbft/commits/tmp/merge-message-interfaces) passes the [CI test](https://github.com/Naoya-Horiguchi/minbft/commit/e84074deb7f5d4ddf4bea10f563343b76e778528/checks?check_suite_id=359583163).

core/message-handling.go
sergefdrv: From the perspective of Go type system, those statements refer to different method, although with the same signature. We would need to cast `*messages.Prepare`/`*messages.Commit` to an interface type to make it more generic. But, as you see, the next commit changes this anyway.

sergefdrv:
> do we have a nice tool to support refactoring that has been done in the last commit? I guess changing a struct to an interface is a common refactoring method and this cause tedious and extensive code modification. It is hard to do that manually.

@ynamiki I think [`gofmt`](https://golang.org/cmd/gofmt/) could help with this job, but this time I actually followed compiler errors and fixed most of them with regex substitution...

sergefdrv:
> I tried to resolve conflict, but I found that I had to change core/request_test.go (not listed in the conflicting files) to pass the unit tests, 

Maybe we better first finalize and merge #142, then rebase this PR.

> so I think we can't simply use Web UI and we might have to push origin/master via CLI. 

I am not sure how good the Web UI is for this kind of job :slightly_smiling_face:

> Should I do this? The resolved branch passes the CI test.

No worries, I'll take care of this PR.

core/message-handling.go
ynamiki: Thanks for your explanation.

ynamiki:
> do we have a nice tool to support refactoring that has been done in the last commit?

Thanks to both of you for reply. I understand there are not a one-shot method and we need to update code manually referencing compiler errors.

sergefdrv:
Summary of [changes](https://github.com/hyperledger-labs/minbft/compare/f11add8606d4b56d9f0331e1165a73c8981b3d4d..8984c84d59ff2654a4bf2ad7ef2f68fed310ca1e):

- Rebase with conflict resolution

PR #123, opened 2019-10-15
Detect skipped Commit messages

sergefdrv:
Correct replicas must always generate their Commit messages
sequentially, as they process Prepare messages from the primary.
However, Byzantine replicas might deviate from this behavior and do
not generate Commit messages for some Prepare messages. Current
implementation does not properly detect this.

Due to asynchronous nature of communication, different replicas might
collect different quorums of Prepare/Commit messages, including those
generated by faulty replicas. Thus, in current implementation, some
correct replicas might happen to erroneously skip some prepared
requests following Byzantine replicas. Fix this problem by detecting
and discarding such inconsistently generated Commit messages.

### DIFF
--- a/core/commit.go
+++ b/core/commit.go
@@ -132,6 +132,7 @@ func makeCommitmentCounter(f uint32) commitmentCounter {
 	)
 
 	return func(replicaID uint32, prepare *messages.Prepare) (done bool, err error) {
+		primaryID := prepare.ReplicaID()
 		prepareUI, err := parseMessageUI(prepare)
 		if err != nil {
 			panic(err)
@@ -147,22 +148,34 @@ func makeCommitmentCounter(f uint32) commitmentCounter {
 
 		replicasCommitted := prepareStates[prepareCV]
 		if replicasCommitted == nil {
-			replicasCommitted = make(replicasCommittedMap)
+			replicasCommitted = replicasCommittedMap{
+				primaryID: true,
+			}
 			prepareStates[prepareCV] = replicasCommitted
 		}
 
-		if replicasCommitted[replicaID] {
-			return false, fmt.Errorf("Duplicated commitment")
+		if replicaID != primaryID {
+			for cv := lastDoneCV + 1; cv < prepareCV; cv++ {
+				s := prepareStates[cv]
+				if s != nil && !s[replicaID] {
+					return false, fmt.Errorf("Skipped commitment")
+				}
+			}
+
+			if replicasCommitted[replicaID] {
+				return false, fmt.Errorf("Duplicated commitment")
+			}
+
+			replicasCommitted[replicaID] = true
 		}
-		replicasCommitted[replicaID] = true
 
-		if len(replicasCommitted) == int(f+1) {
-			delete(prepareStates, prepareCV)
-			lastDoneCV++
-
-			return true, nil
+		if len(replicasCommitted) <= int(f) {
+			return false, nil
 		}
 
-		return false, nil
+		lastDoneCV = prepareCV
+		delete(prepareStates, prepareCV)
+
+		return true, nil
 	}
 }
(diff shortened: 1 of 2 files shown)


core/commit.go
```diff
@@ -154,15 +154,23 @@ func makeCommitmentCounter(f uint32) commitmentCounter {
...
-			return true, nil
+		if len(replicasCommitted) < int(f) {
+			replicasCommitted[replicaID] = true
+
+			return false, nil
 		}
 
-		return false, nil
+		lastDoneCV = prepareCV
+		delete(prepareStates, prepareCV)
```
ynamiki, line 177: Do we also delete CVs between lastDoneCV and prepareCV - 1, in case of prepareCV has increased more than one?

core/commit.go
```diff
@@ -154,15 +154,23 @@ func makeCommitmentCounter(f uint32) commitmentCounter {
...
 			return false, fmt.Errorf("Duplicated commitment")
 		}
-		replicasCommitted[replicaID] = true
 
-		if len(replicasCommitted) == int(f+1) {
-			delete(prepareStates, prepareCV)
-			lastDoneCV++
+		for cv := lastDoneCV + 1; cv < prepareCV; cv++ {
+			s := prepareStates[cv]
+			if len(s) != 0 && !s[replicaID] {
```
nhoriguchi: Do we have to check `len(s) != 0` or simply non-nil check is enough?

core/commit.go
```diff
@@ -154,15 +154,23 @@ func makeCommitmentCounter(f uint32) commitmentCounter {
...
 
-			return true, nil
+		if len(replicasCommitted) < int(f) {
+			replicasCommitted[replicaID] = true
+
+			return false, nil
 		}
 
-		return false, nil
+		lastDoneCV = prepareCV
```
nhoriguchi, line 176: This implicitly allows us to increase `lastDoneCV` by more than 1. If that should be considered as error in any possible case, putting some assertion to detect that might be desirable.

sergefdrv: If not the CHECKPOINT message which also gets UI assigned, the PREPARE/COMMIT messages would always have sequential UIs. So, according to the paper, this is not an error.

core/commit.go
sergefdrv: This commit should make it impossible. Because it does not allow to skip commitments, the threshold of `f+1` commitment will be reached without skipping any `prepareCV`.

sergefdrv: We can also double-check this, and panic in case the check fails.

core/commit.go
sergefdrv: I think it has to be like this, because it gets assigned with an empty map at [line 151](https://github.com/hyperledger-labs/minbft/pull/123/files/9d6719f9ab366af1f1b970ecfd7d7ca049b83885#diff-7377bd6b40f03b94c9d00ac162e96100L151).

sergefdrv: However, the first commitment should be for a PREPARE from the primary. I need to give it a though... Thanks for noticing :wink:

nhoriguchi (approved):
Thanks for elaborating, then I'm fine to the proposed change.

ynamiki (approved):

sergefdrv:
@Naoya-Horiguchi @ynamiki Thanks for very useful feedback! Since the changed code relies on the primary's commitment counted first, I adjusted the code to fulfil this assumption on spot.

sergefdrv:
Summary of [changes](https://github.com/hyperledger-labs/minbft/compare/9d6719f9ab366af1f1b970ecfd7d7ca049b83885..ae386e0c505fe181625d5ce41592ee418021fac2):

* Always count primary commitment before ones from backups
* Test for nil map instead of zero length
* Updated commit message

ynamiki (approved):
Thanks for your improvement. It becomes more easy-to-readable code for me.

nhoriguchi (approved):

PR #124, opened 2019-10-28
Support view change in commitment counter

sergefdrv:
This pull request starts adding view change support. The first component to change is `commitmentCounter`. Please see commit messages for more details.

### DIFF
--- a/core/commit.go
+++ b/core/commit.go
@@ -125,6 +125,13 @@ func makeCommitmentCounter(f uint32) commitmentCounter {
 	var (
 		lock sync.Mutex
 
+		// Current view number
+		view uint64
+
+		// UI counter of the first Prepare in the view
+		firstCV uint64 = 1
+
+		// Primary UI counter of the last quorum
 		lastDoneCV uint64
 
 		// Primary UI counter -> replicasCommittedMap
@@ -133,6 +140,7 @@ func makeCommitmentCounter(f uint32) commitmentCounter {
 
 	return func(replicaID uint32, prepare *messages.Prepare) (done bool, err error) {
 		primaryID := prepare.ReplicaID()
+		prepareView := prepare.View()
 		prepareUI, err := parseMessageUI(prepare)
 		if err != nil {
 			panic(err)
@@ -141,6 +149,15 @@ func makeCommitmentCounter(f uint32) commitmentCounter {
 
 		lock.Lock()
 		defer lock.Unlock()
+
+		if prepareView < view {
+			return false, nil
+		} else if prepareView > view {
+			view = prepareView
+			firstCV = prepareCV
+			lastDoneCV = 0
+			prepareStates = make(map[uint64]replicasCommittedMap)
+		}
 
 		if prepareCV <= lastDoneCV {
 			return true, nil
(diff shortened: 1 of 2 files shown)


core/commit.go
```diff
@@ -142,6 +150,15 @@ func makeCommitmentCounter(f uint32) commitmentCounter {
 		lock.Lock()
 		defer lock.Unlock()
 
+		if prepareView < view {
+			return false, nil
```
nhoriguchi, line 154: <del>Returning an error string might be useful as other error paths do. </del> Nop, this case might be one that should be silently ignored. I withdraw this comment, sorry.

core/commit.go
```diff
@@ -142,6 +150,15 @@ func makeCommitmentCounter(f uint32) commitmentCounter {
...
+		} else if prepareView > view {
+			view = prepareView
+			firstCV = prepareCV
+			lastDoneCV = 0
```
nhoriguchi, line 158: I think that `lastDoneCV` should be turned back to the last checkpoint value, but checkpoint is not implemented yet so setting to 0 now. Maybe that's fine.

core/commit.go
```diff
@@ -133,6 +140,7 @@ func makeCommitmentCounter(f uint32) commitmentCounter {
 
 	return func(replicaID uint32, prepare *messages.Prepare) (done bool, err error) {
 		primaryID := prepare.ReplicaID()
+		prepareView := prepare.View()
```
nhoriguchi, line 143: `primaryID` is always equal to `prepareView % n`, but if that's not the case, the message should be suspected seriously. So is it worth asserting that?

core/commit_test.go
```diff
@@ -237,6 +238,40 @@ func TestMakeCommitmentCounter(t *testing.T) {
 			replicaID: 2,
 			ok:        false,
 			done:      false,
+		}, {
+			desc:      "Fist commitment in a new view",
```
nhoriguchi: s/Fist/First/ ?
Ditto line 269.

core/commit.go
```diff
@@ -125,6 +125,13 @@ func makeCommitmentCounter(f uint32) commitmentCounter {
 	var (
 		lock sync.Mutex
 
+		// Current view number
+		view uint64
```
ynamiki, line 129: At line 153, this variable is used without explicit initialization. Do we add ` = 0` here to clarify we expect zero for this?

core/commit.go
sergefdrv: The only purpose of `lastDoneCV` is to limit `prepareStates` in size. In fact, `lastDoneCV` is the highest counter value that is known to have reached the quorum. Zero is not a valid counter value, so it indicates that no quorum has been reached in the view.

core/commit.go
sergefdrv: As documented in definition of `commitmentCounter` type, the passed Prepare message is supposed to be validated for internal consistency. This means `prepare.ReplicaID()` must denote the primary replica of view `prepare.View()` in the current configuration.

sergefdrv:
Summary of [changes](https://github.com/hyperledger-labs/minbft/compare/2ee32e02548becd6590b258c713c3f19fba8891b..0a72bb25260e6e8ace4b5ec902b4a93b032ccdeb):

-  Fixed typos

core/commit.go
sergefdrv: It is not really necessary. Unlike C/C++, Go variables are always initialized with [a default value](https://golang.org/ref/spec#The_zero_value), which is `0` for numeric types.

core/commit.go
nhoriguchi: Okay, thanks for clarification.

nhoriguchi (approved):
The proposed change looks fine to me. Thanks.

core/commit.go
ynamiki: Yes, however I tend to explicitly initialize a variable if it is used without assignment. It's my coding preference so no problem.

ynamiki (approved):
Looks good to me, thanks.

PR #127, opened 2019-11-22
Introduce prepare timer

nhoriguchi:
This pull request suggests to introduce prepare timer based on the discussion #5.  The outline is described mainly in [the 1st patch](https://github.com/Naoya-Horiguchi/minbft/commit/588a5ebb1681df296879137a7fa1eae701fe3e23).  But I admit that I'm still on the way and there're some points to discuss/solve:

- Using `generatedMessageConsumer` to forward `REQUEST` messages: I just used existing code for the current purpose, but it seems not comply with the current abstraction rule.  I guess that we need have `prepareTimeoutHandler`, but that's not done in this version.
- Adding some "implement" method on type `messages.Request`: this is to make the type implement interface `messages.ReplicaMessage`. I feel that this change has some impact and might need more adjustment on `messages/messages.go`.
- Fault simulator: I do some testing by modifying the sample application to do some faulty behavior, I think this might be helpful to merge it into mainline code as "faulty mode", but maybe that need more consideration on design.
- To refactor duplicated code between request timer and prepare timer: I copied most of prepare timer from request timer, so there's lines of duplicate code, I tried to define common type/interface and defining two timers in hierarchal manner, but I still don't make it.

I'm glad if I get some feedbacks.

CircleCI status is [here](https://circleci.com/gh/Naoya-Horiguchi/minbft/77).


sample/peer/cmd/request.go
```diff
@@ -57,19 +59,43 @@ func init() {
...
+		ctx, cancel = context.WithTimeout(ctx, timeout)
+		time.AfterFunc(timeout, cancel)
+	}
+
+	go requestInternal(ctx, ch, client, req)
+	select {
+	case <-ctx.Done():
+		fmt.Println("Client Request timer expired.")
+	case <-ch:
+	}
```
sergefdrv: I wonder if this can be simply:

```go
var timeoutChan <-chan time.Time

if timeout > 0 {
  timeoutChan = time.After(timeout)
}

select {
case res := <-client.Request(req):
  fmt.Println("Reply:", string(res))
case <-timeoutChan:
  fmt.Println("Client Request timer expired.")
  os.Exit(1)
}
```

core/internal/clientstate/client-state.go
```diff
@@ -151,10 +164,20 @@ func WithRequestTimeout(timeout func() time.Duration) Option {
 	}
 }
 
+// WithPrepareTimeout specifies a function that returns the duration
+// to use when starting a new prepare timeout timer. Zero or negative
+// duration disables the timeout. The timeout is disabled by default.
+func WithPrepareTimeout(timeout func() time.Duration) Option {
```
sergefdrv: I am not sure anymore if it was a good idea (of mine) to supply request timeout provider as optional parameter, with timeout disabled by default. Maybe it would better to pass the timeout providers as mandatory parameters of `New`?

core/internal/clientstate/prepare-timeout.go
```diff
@@ -0,0 +1,63 @@
...
+
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/hyperledger-labs/minbft/core/internal/timer"
+)
+
+type prepareTimerState struct {
```
sergefdrv: I think we should better generalize `requestTimerState` into `timerState` that could serve for both request and prepare timer.

core/request.go
```diff
@@ -305,3 +315,32 @@ func makeRequestTimeoutProvider(config api.Configer) requestTimeoutProvider {
...
+			logger.Infof("Prepare timer expired: client=%d view=%d", clientID, view)
+			consume(request)
+		})
+	}
+}
+
+// makePrepareTimerStopper constructs an instance of
+// prepareTimerStopper.
+func makePrepareTimerStopper(provideClientState clientstate.Provider) prepareTimerStopper {
+	return func(clientID uint32) {
```
sergefdrv, line 344: This deviates from description of `prepareTimerStopper` where the parameter is a replica identifier. Should we better use client identifier?

sergefdrv (changes requested):
Thanks for taking the initiative to solve the issue. This doesn't look an easy one. Perhaps the most difficult part is the actual mechanism to relay REQUEST message. Therefore, I would suggest to split the work. Maybe first focus on handling the timeout itself, and just report the timeout event in the diagnostic message log. Then we could tackle message relaying as a separate issue.

I was not sure about changing the request timeout semantics. Originally, it was the maximum expected time for a received request to become ready for execution. This PR delays starting the request timer until the PREPARE message is received. However, when a correct replica has accepted a PREPARE message, it will effectively multicast it to other replicas in COMMIT message. This ensures eventual request execution unless view change is triggered. In other words, once a single correct replica has accepted a PREPARE message, reaching the consensus does not depend on the primary anymore. Thus, the request timer started at that point would not really serve for faulty primary detection.

I suggest we start both prepare and request timer upon reception of a new REQUEST message. If the replica does not accept the corresponding PREPARE message for longer than the prepare timeout, it would first assume the client was faulty. Then it will relay the REQUEST message to the primary and restart the request timer. I am currently discussing this scheme with experts, but I am not sure how soon I can get their feedback.

> * Using `generatedMessageConsumer` to forward `REQUEST` messages: I just used existing code for the current purpose, but it seems not comply with the current abstraction rule.  I guess that we need have `prepareTimeoutHandler`, but that's not done in this version.
>
> * Adding some "implement" method on type `messages.Request`: this is to make the type implement interface `messages.ReplicaMessage`. I feel that this change has some impact and might need more adjustment on `messages/messages.go`.

Those points need more consideration and there is clearly something to be changed. It would not be appropriate to treat REQUEST as replica message. When a prepare timeout expires, the REQUEST message should only be forwarded to the primary, there is no need to multicast it to all replicas. Maybe we should open a dedicated message stream to the primary for REQUEST message forwarding. 

Another consideration is cancelling of the transmission. If a new REQUEST is accepted from the same client, forwarding of the old REQUEST should be cancelled. At least, there should be a clear limit for REQUEST messages being forwarded, they should not pile up indefinitely.

> * Fault simulator: I do some testing by modifying the sample application to do some faulty behavior, I think this might be helpful to merge it into mainline code as "faulty mode", but maybe that need more consideration on design.

As opposed to incorporating faulty behavior into normal code, we can create some kind of tool for testing. The tool would act as a client or replica following some predefined scenario, including Byzantine behavior. What do you think?

I think we could continue reviewing and merge the commit 2006f59bf7b3b51a6abf1896a9e15d0d8329cb53 in a separate PR.

sample/peer/cmd/request.go
nhoriguchi: I used `context` to realize "never expire if timeout is not specified", but simple is better so I'll try with your code, thank you.

core/internal/clientstate/client-state.go
nhoriguchi: > Maybe it would better to pass the timeout providers as mandatory parameters of New?

OK, I'll do this.

core/request.go
nhoriguchi: Ah, that's right, I'll fix it.

core/internal/clientstate/prepare-timeout.go
nhoriguchi: Yes, that will be done next.

nhoriguchi:
> ... Therefore, I would suggest to split the work. Maybe first focus on handling the timeout itself, and just report the timeout event in the diagnostic message log. Then we could tackle message relaying as a separate issue.

OK.

> 
> I was not sure about changing the request timeout semantics. Originally, it was the maximum expected time for a received request to become ready for execution. This PR delays starting the request timer until the PREPARE message is received. However, when a correct replica has accepted a PREPARE message, it will effectively multicast it to other replicas in COMMIT message. This ensures eventual request execution unless view change is triggered. In other words, once a single correct replica has accepted a PREPARE message, reaching the consensus does not depend on the primary anymore. Thus, the request timer started at that point would not really serve for faulty primary detection.
> 
> I suggest we start both prepare and request timer upon reception of a new REQUEST message. If the replica does not accept the corresponding PREPARE message for longer than the prepare timeout, it would first assume the client was faulty. Then it will relay the REQUEST message to the primary and restart the request timer. I am currently discussing this scheme with experts, but I am not sure how soon I can get their feedback.

Thanks for detailed explanation, this note is very important for me. I'll change request timer behavior as you stated above.

...
> Another consideration is cancelling of the transmission. If a new REQUEST is accepted from the same client, forwarding of the old REQUEST should be cancelled. At least, there should be a clear limit for REQUEST messages being forwarded, they should not pile up indefinitely.

In the current version, a backup replica "forwards" (i.e. multicasts) a REQUEST message only once when the prepare timer expires, so maybe no worry about request's pile-up. And if a replica receives the old REQUEST message after the new REQUEST messages is accepted, the old one should be properly rejected by `CaptureRequestSeq` mechanism.

> > * Fault simulator: I do some testing by modifying the sample application to do some faulty behavior, I think this might be helpful to merge it into mainline code as "faulty mode", but maybe that need more consideration on design.
> 
> As opposed to incorporating faulty behavior into normal code, we can create some kind of tool for testing. The tool would act as a client or replica following some predefined scenario, including Byzantine behavior. What do you think?

Any faulty behaviors on client side can be mimicked by testing tools. As for replica side, some external attacks (like replay attack) can be simulated by tools, but we might need some internal module for testing inside attacks. The internal module should be switched on or off on compile time to ensure that it's never used on production setting.

> I think we could continue reviewing and merge the commit [2006f59](https://github.com/hyperledger-labs/minbft/commit/2006f59bf7b3b51a6abf1896a9e15d0d8329cb53) in a separate PR.

OK, I'll separate the suggestion for this commit too.

sergefdrv:
> In the current version, a backup replica "forwards" (i.e. multicasts) a REQUEST message only once when the prepare timer expires, so maybe no worry about request's pile-up.

If I understood correctly, the prepare timer is started for each new REQUEST message. When it expires, the message is appended to the message log. My concern was that there is no limit for the number of REQUEST messages from the same client that can be appended to the message log.

> And if a replica receives the old REQUEST message after the new REQUEST messages is accepted, the old one should be properly rejected by `CaptureRequestSeq` mechanism.

That is correct, I was just thinking of wasted network bandwidth.

> [...] we might need some internal module for testing inside attacks. The internal module should be switched on or off on compile time to ensure that it's never used on production setting.

I was not quite clear about what you mean by "inside attacks". Do you mean simulation of intrusion attacks on replica's untrusted code? Such compromised replica can be simply considered faulty. And the fraction of faulty replicas is assumed limited. What is your perspective on this?

sergefdrv:
> The internal module should be switched on or off on compile time to ensure that it's never used on production setting.

My biggest concern with this approach is code clarity. I'm afraid embedding faulty logic for testing can make the code hard to follow.

nhoriguchi:
> > [...] we might need some internal module for testing inside attacks. The internal module should be switched on or off on compile time to ensure that it's never used on production setting.
> 
> I was not quite clear about what you mean by "inside attacks". Do you mean simulation of intrusion attacks on replica's untrusted code?

Yes, I have that in mind.

> Such compromised replica can be simply considered faulty. And the fraction of faulty replicas is assumed limited. What is your perspective on this?

What I thought was that MinBFT code is public, so attackers (note that some legitimate member could be an attacker on trustless network) can learn it and easily embed any compromised behavior (which might be hard for correct replicas to find). So I thought that we will be better to provide working demonstration code of typical scenarios with byzantine faulty replicas for anyone to verify.

> > The internal module should be switched on or off on compile time to ensure that it's never used on production setting.
> 
> My biggest concern with this approach is code clarity. I'm afraid embedding faulty logic for testing can make the code hard to follow.

Yes, I understand that, so I'm ok to start with separate cod/branch.

Anyway, thank you for the idea.

sergefdrv:
> What I thought was that MinBFT code is public, so attackers (note that some legitimate member could be an attacker on trustless network) can learn it and easily embed any compromised behavior (which might be hard for correct replicas to find). So I thought that we will be better to provide working demonstration code of typical scenarios with byzantine faulty replicas for anyone to verify.

So you consider simulating a tolerated fraction of Byzantine replicas? If so, I thought we might better have some dummy replica nodes that behave Byzantine, following some predefined scenario. 

What you were probably concerned with was that the replica behavior we would like to test might be quite complicated. In that case, we might need some smarter tool, which behaves almost the same as correct replica.

I think we can start exploring this issue with a scenario-based tool and see how it goes.

I can think of two different approaches with the tool: (1) the tool mimics an individual replica's behavior, and (2) the tool mimics the behavior of all replicas in the network, except the tested one. In (1), it would be easier to define the Byzantine scenario for the simulated faulty replica. In (2), it would be easier to define the scenario, taking asynchronous message exchange into account. Maybe there are more possibilities, or it might make sense to use a hybrid approach.

nhoriguchi:
> So you consider simulating a tolerated fraction of Byzantine replicas?

Yes, in most severe cases, `f` replicas are in conspiracy to disturb consensus process.

> If so, I thought we might better have some dummy replica nodes that behave Byzantine, following some predefined scenario.
> 
> What you were probably concerned with was that the replica behavior we would like to test might be quite complicated. In that case, we might need some smarter tool, which behaves almost the same as correct replica.
> 
> I think we can start exploring this issue with a scenario-based tool and see how it goes.
> 
> I can think of two different approaches with the tool: (1) the tool mimics an individual replica's behavior, and (2) the tool mimics the behavior of all replicas in the network, except the tested one. In (1), it would be easier to define the Byzantine scenario for the simulated faulty replica. In (2), it would be easier to define the scenario, taking asynchronous message exchange into account. Maybe there are more possibilities, or it might make sense to use a hybrid approach.

The hybrid approach sounds nice to me, because these approaches complement each other: (1) is for integration testing which should define expected behaviors as a system for given scenarios, so that's close to what I imaged, but (2) is also important because it tests whether a correct replica behave as expected for given scenarios.

ynamiki (commented):
As Sergey has wrote, I think it is better to have `timerState` struct. Sorry for my late response.

nhoriguchi:
Summary of changes, the latest version:
- introduces `timerState struct`,
- proposes only prepare timer implementation, and message forwarding is not included (#140), and
- postpones to make timeout parameter mandatory (#139)

core/internal/clientstate/prepare-timeout.go
ynamiki: May I ask you why we define both `requestTimerState` and `prepareTimerState` struct? They have same definition, so I think we could define two variable for request and prepare timer state of single `timerState` type.

nhoriguchi: No technical reason actually. Maybe I misread the feedback, and now I see that I can completely remove `requestTimerState` and `prepareTimerState`. I'll try it.

core/internal/clientstate/client-state.go
sergefdrv: To be solved in #139.

nhoriguchi:
I updated the branch again. In this version I tried two things:

- timeout duration is passed to client-state provider as a mandatory parameter, and
- more code generalization (so that there's no longer `requestTimerState`)

core/internal/clientstate/request-timeout.go
```diff
@@ -24,22 +24,26 @@ import (
...
 	sync.Mutex
 
-	// Request timeout timer
-	requestTimer timer.Timer
+	timerProvider timer.Provider
 
-	opts *options
+	// Request timeout timer
+	requestTimer   timer.Timer
+	requestTimeout time.Duration
```
sergefdrv: Nitpicking: Maybe move `requestTimeout` filed definition on the next line after `timerProvider` so that both configuration parameters are close together?

core/request.go
```diff
@@ -122,9 +121,6 @@ type requestTimerStopper func(clientID uint32)
 // started. It is safe to invoke concurrently.
 type requestTimeoutHandler func(view uint64)
 
-// requestTimeoutProvider returns current request timeout duration.
-type requestTimeoutProvider func() time.Duration
```
sergefdrv, line 126: I see you are changing request timeout to a constant value as opposed to a dynamically retrieved value. I think it is a fair change, but let me just expand a bit on this, since this is an important detail.

In MinBFT paper, it is only the view change timeout that gets doubled on each attempt to establish a stable view. That is to accommodate for possibly increased communication delay. However, [the supplementary material](http://www.di.fc.ul.pt/~bessani/publications/tc11-minimal-supp.pdf) (the proof) also mentions doubling request timeout value. That was the reason why we originally made the request timeout value dynamically retrieved here.

In the end, this is related to different kinds of [partially synchronous communication model](https://groups.csail.mit.edu/tds/papers/Lynch/jacm88.pdf). One model makes no assumption on communication delay, except that it doesn't grow indefinitely. Another model assumes that there might be periods of asynchrony with unknown communication delay, but eventually the system reaches a so-called global synchronization time after which the communication delay stays within some known bound (for some time). 

As a first step, we can assume the model with global synchronization time, and keep the request/prepare timeout value constant.

core/internal/clientstate/client-state.go
```diff
@@ -116,7 +126,7 @@ func New(opts ...Option) State {
 
 	s.seqState = newSeqState()
 	s.replyState = newReplyState()
-	s.requestTimerState = newRequestTimeoutState(&s.opts)
+	s.timerState = newTimeoutState(s.opts.timerProvider, requestTimeout, prepareTimeout)
```
sergefdrv: I think what @ynamiki actually [meant](https://github.com/hyperledger-labs/minbft/pull/127#discussion_r355902769), and I tend to agree, was to have two fields in `clientState`: one for request and another for prepare timer. The timer implementation would be the same (`timerState`), but initialized with different timeout values. Not sure, where would be better to translate `State` methods to respective `timerState` methods though. Maybe something like this:

```go
type requestTimer timerState

func newRequestTimer(provideTimer timer.Provider, timeout time.Duration) *requestTimer {
	return newTimerState(provideTimer, timeout)
}

func (t *requestTimer) StartRequestTimer(handleTimeout func()) {
	t.StartTimer(handleTimeout)
}

func (t *requestTimer) StopRequestTimer() {
	t.StopTimer()
}
```

```go
type prepareTimer timerState

func newPrepareTimer(provideTimer timer.Provider, timeout time.Duration) *prepareTimer {
	return newTimerState(provideTimer, timeout)
}

func (t *prepareTimer) StartPrepareTimer(handleTimeout func()) {
	t.StartTimer(handleTimeout)
}

func (t *prepareTimer) StopPrepareTimer() {
	t.StopTimer()
}
```

Alternatively, we can translate the methods in `client-state.go`.

core/prepare.go
```diff
@@ -64,12 +64,16 @@ func makePrepareValidator(n uint32, verifyUI uiVerifier, validateRequest request
...
+func makePrepareApplier(id uint32, prepareSeq requestSeqPreparer, collectCommitment commitmentCollector, handleGeneratedUIMessage generatedUIMessageHandler, startReqTimer requestTimerStarter, stopPrepTimer prepareTimerStopper) prepareApplier {
 	return func(prepare *messages.Prepare) error {
-		if new := prepareSeq(prepare.Msg.Request); !new {
+		request := prepare.Msg.Request
+		if new := prepareSeq(request); !new {
 			return fmt.Errorf("Request already prepared")
 		}
 
+		stopPrepTimer(prepare.Msg.ReplicaId)
+		startReqTimer(request.Msg.ClientId, prepare.Msg.View)
```
sergefdrv: The request timer is started by `requestApplier`, I think there is no need to restart the request timer here.

sergefdrv (changes requested):
Thanks for your effort, Naoya, but I think the code needs some more adjustment.

core/internal/clientstate/request-timeout.go
nhoriguchi: Applied to the latest patch, thank you.

core/internal/clientstate/client-state.go
ynamiki: Thanks @sergefdrv for clarifying the topic. That is what I should have written.

Actually, I thought like as follows, but now I find that this doesn't conform to the definition of `clientState` which use the embedding.

```go
type clientState struct {
	*seqState
	*replyState
	requestTimerState timerState
	prepareTimerState timerState
	...
}
```

core/request.go
nhoriguchi: After applying all comments except this, I found that making timeout optional and making timeout constant are independent of each other, so I shouldn't have made timeout value constant. And keeping timeout dynamic makes the needed change simpler, so I happily take the approach.

And thank you for explaining the background of the current design, I agree with it.

core/internal/clientstate/client-state.go
nhoriguchi: I might slightly turn away from the above advice, but I added two fields of type `*timerState` to `clientState` struct. This change requires me to add some interface functions like `StartRequestTimer` directly to `clientState`, because `timerState` is not embedded in it any more. This might not be perfect, so if you have better ideas, feel free to share them.

core/prepare.go
nhoriguchi: Applied, thank you.

core/internal/clientstate/client-state.go
sergefdrv: I think this is a good compromise :+1:

sergefdrv (changes requested):
Maybe also extend `core/internal/clientstate/timeout_test.go` so that it tests prepare timer methods, or just test `timerState` and assume its methods are invoked correctly from the methods of `clientState`?

nhoriguchi:
> Maybe also extend `core/internal/clientstate/timeout_test.go` so that it tests prepare timer methods, or just test `timerState` and assume its methods are invoked correctly from the methods of `clientState`?

Pushed one additional commit, thank you.

sergefdrv (approved):

sergefdrv:
@Naoya-Horiguchi Looks good. Thanks!

PR #129, opened 2019-11-26
CI with GitHub Actions

### LINKED ISSUE #128: Try GitHub Actions
GitHub Actions is now available. Even we don't have a strong motivation to move from our current CI (CircleCI), I would like to check the features and try the new service.

https://help.github.com/en/actions

ynamiki:
[GitHub Actions](https://help.github.com/en/actions) is now available and now we have another alternative for CI.

This pull request activates CI with GitHub Actions. We will try this GitHub Actions in addition to the current CircleCI for some weeks then decide which one to use.

At this time, I prefer GitHub Actions:
- Set up: GitHub Action has already integrated, nothing to do. CircleCI needs some processes to integrate the service to GitHub. I spent much time to this part for the initial setup and the trouble shooting (the trigger worked only for me). Note that we don't have Admin privilege, so we need to request assistance to the LF team in some case.
- Configuration: There are small difference, but both are easy to write.
- Environment: GitHub Actions provides a VM while CircleCI provides a container to run CI. I don't find any good/bad effects from this difference.

This PR resolves #128.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,31 @@
+name: Continuous integration
+on: push
+env:
+  SGX_SDK_URL: https://download.01.org/intel-sgx/linux-2.3.1/ubuntu18.04/sgx_linux_x64_sdk_2.3.101.46683.bin
+  GOLANGCI_LINT_URL: https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh
+  GOLANGCI_LINT_VERSION: v1.21.0
+  SGX_MODE: SIM
+
+jobs:
+  build:
+    name: Continuous integration
+    runs-on: ubuntu-18.04
+    steps:
+      - name: Install Intel SGX SDK
+        run: |
+          curl -s $SGX_SDK_URL -o sgx_linux_x64_sdk.bin
+          chmod +x sgx_linux_x64_sdk.bin
+          echo -e "no\n/opt/intel" | sudo ./sgx_linux_x64_sdk.bin
+          rm sgx_linux_x64_sdk.bin
+      - name: Install GolangCI-Lint
+        run: |
+          curl -sfL $GOLANGCI_LINT_URL | sudo sh -s -- -b /usr/local/bin $GOLANGCI_LINT_VERSION
+      - uses: actions/checkout@v1
+      - name: Test
+        run: |
+          source /opt/intel/sgxsdk/environment
+          LD_LIBRARY_PATH="$PWD/usig/sgx/shim:$LD_LIBRARY_PATH" \
+          make build check
+      - name: Lint
+        run: |
+          make lint


ynamiki:
We have an error in the GitHub Actions (and this is not occurred in CircleCI):

```
golangci-lint run ./...
usig/sgx/usig-enclave.go:260:145: SA4000: identical expressions on the left and right side of the '==' operator (staticcheck)
	sgxErr := C.call_usig_seal_key(function, enclave.enclaveID, &sealedData, &sealedDataSize)
```

https://github.com/ynamiki/minbft/commit/6149de338790296ff86d2c3ca08f54eb2ff6649d/checks?check_suite_id=328572416

I think this is a false positive case reported in https://github.com/dominikh/go-tools/issues/430 and have a relationship with Go. Currently, we use Go 1.12.7, which is provided as default in the GitHub Actions environment. This is different from the version used in CircleCI: 1.11.4. I will try update the configuration to use the same version of Go and check results.

sergefdrv:
> I think this is a false positive case reported in [dominikh/go-tools#430](https://github.com/dominikh/go-tools/issues/430) and have a relationship with Go. Currently, we use Go 1.12.7, which is provided as default in the GitHub Actions environment. This is different from the version used in CircleCI: 1.11.4. 

This seems to be linting issue. Namely, false positive of `staticcheck`. It should be fixed in `staticcheck` as of 2019.2.1. Maybe upgrading to the latest `GolangCI-Lint` will help?

sergefdrv:
@ynamiki Thanks for doing this :+1: The configuration looks fine. Let's fix the linting issue and try it out.

ynamiki:
The linter error has fix by upgrading golangci-lint. Thanks for the comment.

ynamiki:
At this time, this commit will not replace the current CI by CircleCI. Let's disscuss some weeks later whether switch to this or not after trying the new CI system.

sergefdrv (approved):
Look good :+1: Thanks!

We would need to update the required version of Go in our ReadMe if we decide to switch to GitHub actions.

nhoriguchi (approved):
Thanks, I'll merge this soon, then let's see what happens :+1:

.github/workflows/main.yml
```diff
@@ -0,0 +1,31 @@
+name: Continuous integration
+on: push
```
nhoriguchi, line 2: (Sorry, I should have noticed this before pushing merge button...) if we'd like to trigger a job on creating pull request, writing like `on: [push, pull_request]` here might be better?

ynamiki: I think it could be enough with only `push` but I will add `pull_request` if the action is not triggered by a next new PR. Thanks.

https://help.github.com/en/actions/automating-your-workflow-with-github-actions/events-that-trigger-workflows

sergefdrv: Apparently, Naoya was right. My latest pull request (#131) did not trigger CI action in this repository. It was only triggered in my fork.

ynamiki: I have updated the configuration in #133. Thanks!

PR #131, opened 2019-11-30
Maintain pending request list

sergefdrv:
This pull request introduces a mechanism to maintain a set of pending requests. A pending request is a request that was processed, but not yet accepted for execution. The pending request set will be used to restart consensus after view change is completed. All replicas will restart timers for all pending requests. In addition, the new primary will prepare any pending request that does not appear in the new-view certificate.


core/internal/requestlist/request-list.go
```diff
@@ -0,0 +1,80 @@
...
+package requestlist
+
+import (
+	"sync"
+
+	"github.com/hyperledger-labs/minbft/messages"
+)
+
+// List defines methods to manipulate a set of Request messages. The
+// set is only capable of keeping a single message per client. All
```
nhoriguchi, line 26: (Might be a stupid question) is "keeping a single message per client" enough? If we have multiple pending requests for a client, the older one seems to be just ignored. If you have a good reason to do it or have a plan to extend it later, that's ok.

sergefdrv: MinBFT paper and current implementation does not allow correct clients to submit multiple requests in parallel. The client has to wait for the last request to complete before submitting another request. Different clients, in contrast, can submit requests independently and asynchronously w.r.t. each other.

We can think of relaxing this requirement as a future optimization. In fact, we have this in [our list of features to consider](https://github.com/hyperledger-labs/minbft#roadmap): "Asynchronous requests: enabling parallel processing of requests".

sergefdrv: However, let me think about this detail a bit more. Now, I'm not sure if storing only the last pending request per client would be enough to use this component in view change...

sergefdrv: So any request that was executed will be present in a valid new-view certificate and will be propagated to the new view by the view change mechanism. That means, after view change is completed, consensus may only need to be restarted for the last client's request. Thus, it seems enough to keep a single message per client in the pending request list.

sergefdrv:
Summary of [changes](https://github.com/hyperledger-labs/minbft/compare/2fed7e9a2ed495206dc52e87c660b964ba52309e..ab6483aaa11634b5b188d01b661ae1cdef2a0e50):

- Rebase with conflict resolution and some required small adjustments

core/internal/requestlist/request-list.go
nhoriguchi: Thanks, so it's clear to me now, and I agree with the proposed change.

nhoriguchi (approved):

sergefdrv:
Sorry, I seem to have mistakenly closed this issue...

nhoriguchi:
After merging this commit, I received [CI failure report](https://github.com/hyperledger-labs/minbft/commit/b6ccd6378f166a02702464ccc0c22569eb1c55d6/checks) which is a known issue #110, so I don't think this pull request is responsible for it.

sergefdrv:
> After merging this commit, I received [CI failure report](https://github.com/hyperledger-labs/minbft/commit/b6ccd6378f166a02702464ccc0c22569eb1c55d6/checks) which is a known issue #110, so I don't think this pull request is responsible for it.

Yes, that was the same issue which I'm currently working on...

PR #132, opened 2019-11-30
Fix race for request execution

### LINKED ISSUE #130: Race condition for request execution
Message streams from different replicas are processed concurrently. Even though PREPARE/COMMIT messages from the same replica are always processed sequentially, there is still a possible race condition for request execution. 

The race can happen because there is no synchronization in `makeCommitmentCollector` between invocations of `countCommitment` and `executeRequest`. It can happen that `countCommitment` returns `true` for two subsequent prepared requests in two goroutines. That is because one goroutine may first supply the last required commitment for one prepared request, whereas another goroutine can supply a commitment for the same prepared request (which is ignored) followed by the last required commitment of the subsequent prepared request. Then two goroutines will race to execute those requests.

sergefdrv:
This pull request resolves #130 according to https://github.com/hyperledger-labs/minbft/issues/130#issuecomment-559488190.

### DIFF
--- a/core/commit.go
+++ b/core/commit.go
@@ -55,7 +55,7 @@ type commitmentCollector func(replicaID uint32, prepare *messages.Prepare) error
 // commitments from different replicas are counted for the supplied
 // Prepare, such that the threshold to execute the prepared operation
 // has been reached. An error is returned if any inconsistency is
-// detected. It is safe to invoke concurrently.
+// detected.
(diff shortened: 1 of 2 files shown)


nhoriguchi (approved):
Looks good to me :+1:

ynamiki (approved):
Thanks!

PR #134, opened 2019-12-02
peer/cmd/request: Introduce request timeout option

nhoriguchi:
This pull request is separated from #127 for client request timeout part as recommended in [comment](https://github.com/hyperledger-labs/minbft/pull/127#pullrequestreview-322880648).

### DIFF
--- a/sample/peer/cmd/request.go
+++ b/sample/peer/cmd/request.go
@@ -21,6 +21,7 @@ import (
 	"bufio"
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/hyperledger-labs/minbft/api"
 	"github.com/hyperledger-labs/minbft/client"
@@ -57,6 +58,8 @@ func init() {
 	requestCmd.Flags().Int("id", 0, "ID of the client")
 	must(viper.BindPFlag("client.id",
 		requestCmd.Flags().Lookup("id")))
+	requestCmd.Flags().String("timeout", "0", "Timeout for the request")
+	must(viper.BindPFlag("client.timeout", requestCmd.Flags().Lookup("timeout")))
 }
 
 type clientStack struct {
@@ -65,9 +68,20 @@ type clientStack struct {
 }
 
 func request(client client.Client, arg string) {
-	req := []byte(arg)
-	res := <-client.Request(req)
-	fmt.Println("Reply:", string(res))
+	var timeoutChan <-chan time.Time
+	timeout := viper.GetDuration("client.timeout")
+
+	if timeout > 0 {
+		timeoutChan = time.After(timeout)
+	}
+
+	select {
+	case res := <-client.Request([]byte(arg)):
+		fmt.Println("Reply:", string(res))
+	case <-timeoutChan:
+		fmt.Println("Client Request timer expired.")
+		os.Exit(1)
+	}
 }
 
 func requests(args []string) ([]byte, error) {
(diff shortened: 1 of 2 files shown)


sample/peer/cmd/request.go
```diff
@@ -65,9 +68,20 @@ type clientStack struct {
 }
 
 func request(client client.Client, arg string) {
-	req := []byte(arg)
-	res := <-client.Request(req)
-	fmt.Println("Reply:", string(res))
+	var timeoutChan <-chan time.Time
+	timeout := viper.GetDuration("client.timeout")
```
ynamiki, line 72: Memo: Errors (giving "zzz" for the parameter, etc.) are ignored. I found some similar usages in the samples so I would assume this is our policy for error handling.

sample/peer/cmd/request.go
```diff
@@ -65,9 +68,20 @@ type clientStack struct {
...
+	if timeout > 0 {
+		timeoutChan = time.After(timeout)
+	}
+
+	select {
+	case res := <-client.Request([]byte(arg)):
+		fmt.Println("Reply:", string(res))
+	case <-timeoutChan:
+		fmt.Println("Client Request timer expired.")
+		os.Exit(1)
```
ynamiki, line 83: Memo: Output an error message to stdout (not stderr) and exit with 1 for an error case are also our conventions.

ynamiki (approved):
I left some comments, but they are intended to clarify our coding conventions.

sample/peer/cmd/request.go
sergefdrv: If the value cannot be parsed as duration (like "zzz"), `GetDuration` returns the default value of `time.Duration` (which is `time.Duration(0)`). If we'd like to detect this better, we could use `viper.Get` and convert the value using [`github.com/spf13/cast.ToDurationE`](https://godoc.org/github.com/spf13/cast#ToDurationE).

sergefdrv: Even though this is just a sample application, this kind of error detection might help avoid surprising results if there is a simple typo in configuration. Maybe address in a separate issue? The same also applies to `sample/config` package.

sample/peer/cmd/request.go
sergefdrv: [Initially, `request` function would return an error](https://github.com/hyperledger-labs/minbft/blob/50d684f2aff06919cbcf40287937a6ef55140063/sample/peer/cmd/request.go#L70) that is propagated by `cobra` up the call stack, and finally reported [here](https://github.com/hyperledger-labs/minbft/blob/50d684f2aff06919cbcf40287937a6ef55140063/sample/peer/cmd/root.go#L48). Maybe we should restore that behavior (in a separate PR)? 

In that case, it seems not necessary to output the error here, because `sample/peer/cmd.Execute` will do that. We might also want to report the error to `stderr` rather than `stdout`.

sergefdrv (approved):
I left some comments regarding error handling which we could solve as separate issues. This PR itself looks good and ready to merge. Thanks!

sergefdrv:
@Naoya-Horiguchi CircleCI seems to have stucked again... Would you mind amending the last commit and force-pushing to trigger the CI? Then this MR can be merged.

sample/peer/cmd/request.go
nhoriguchi: @ynamiki, @sergefdrv thank you for the comments, I fully agree with fixing error checking over the sample application in a separate PR.

nhoriguchi:
I did git-force-push and now CI seems to work well.

PR #142, opened 2019-12-13
Fix prepare/request timers

sergefdrv:
This pull request fixes some issues with prepare timer that were missed when reviewing #127.

Depending on how we decide to solve the issue noted in "WIP: core: Check request ID before stopping request/prepare timer", we might decide to do the opposite of what "RFC: core: Pass client ID to prepareTimerStarter" suggests.

### DIFF
--- a/core/commit.go
+++ b/core/commit.go
@@ -116,7 +116,7 @@ func makeCommitmentCollector(countCommitment commitmentCounter, retireSeq reques
 		}
 
 		pendingReq.Remove(request.Msg.ClientId)
-		stopReqTimer(request.Msg.ClientId)
+		stopReqTimer(request)
 		executeRequest(request)
 
 		return nil
--- a/core/commit_test.go
+++ b/core/commit_test.go
@@ -154,8 +154,8 @@ func TestMakeCommitmentCollector(t *testing.T) {
 		args := mock.MethodCalled("requestSeqRetirer", request)
 		return args.Bool(0)
 	}
-	stopReqTimer := func(clientID uint32) {
-		mock.MethodCalled("requestTimerStopper", clientID)
+	stopReqTimer := func(request *messages.Request) {
+		mock.MethodCalled("requestTimerStopper", request)
 	}
 	executeRequest := func(request *messages.Request) {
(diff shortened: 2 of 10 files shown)


core/commit.go
```diff
@@ -116,6 +116,10 @@ func makeCommitmentCollector(countCommitment commitmentCounter, retireSeq reques
 		}
 
 		pendingReq.Remove(request.Msg.ClientId)
+
+		// FIXME: Only stop the request timer if the request
+		// identifier it was started for matches the one of
+		// the request accepted for execution.
```
Naoya-Horiguchi: I think that when the newer request comes, request timer for the client is restarted by request applier, so the newer request seems not to be prevented by request timer started by the older request? But maybe situation is more complicated by view change...

nhoriguchi: I agree with this, thank you for spotting.

ynamiki: Is there a possibility that a timer is left without stopping forever? I am not sure how a timer is stopped certainly at some point if `stopReqTimer()` doesn't stop the timer at this point.

ynamiki: Is there a possibility that a timer is left without stopping forever? I am not sure how a timer is stopped certainly at some point if `stopReqTimer()` doesn't stop the timer at this point.

sergefdrv: Since correct clients only send their requests sequentially, a replica need to maintain one prepare and one request timer per client. If a newer request is received, the replica can assume the former request was completed and the client timers should be reused. In other words, replicas only maintain timers for the most recent request of each client.

sergefdrv:
Summary of [changes](https://github.com/hyperledger-labs/minbft/compare/506675662e5e1457a0fa92c252adb339eade280b..eeeb504c97fe7ca928416d0dfa8a08146f88faeb):

- Dropped commit "RFC: core: Pass client ID to prepareTimerStarter" and passing Request message to all request/perpare timer abstractions
- Added check for request identifier in timers maintained in client state
- Added unit tests for prepare timer abstractions

nhoriguchi (approved):
Thank you for fixing and improving test coverage.

ynamiki (approved):
Thanks for explanation.

PR #143, opened 2019-12-15
Switch to GitHub Actions

### LINKED ISSUE #120: Upgrade to Go 1.13
Go 1.13 has been [released](https://golang.org/doc/devel/release.html#go1.13). By this release, Go 1.11, which we are using in our CI, is [no longer supported](https://github.com/golang/go/wiki/Go-Release-Cycle#release-maintenance) (i.e. no more security fixes). We should upgrade it.

ynamiki:
We are using GitHub Actions for three weeks and I am comfortable with it. I would like to propose switching our CI system from CircleCI to GitHub Actions.

- GitHub Actions has confirmed to have enough features and performance for our project.
- GitHub Actions is build-in feature so that we don't need to pay attention to the service integration part of GitHub and CircleCI. It is difficult to troubleshoot if we have a problem because we don't have Admin privilege and cannot see details (I had spent a lot of time...).

Currently this PR is set to "Draft". This is because we need to ask the LF IT team to update repository settings (disable CircleCI in the status check) before merge; otherwise we cannot merge any commits after this PR is merged.

Please review the code, and If I get approvals then I will proceed to the setting change and merge.

This patch closes #120 because the new build environment uses Go 1.13.


README.md
```diff
@@ -116,7 +116,7 @@ sudo apt-get install build-essential pkg-config
 
 ### Golang ###
 
-`go1.11` is used to build this project. For installation instructions
+`go1.13` is used to build this project. For installation instructions
```
nhoriguchi: I think that even if new CI uses go1.13, MinBFT should still work with go1.11 or later, so how about describing like "`go1.11` or later is required to build this project. "?

ynamiki: Exactly our (current) codes can works with Go 1.11. However, I think it is difficult to guarantee that while updating the code day by day without a CI. So, how about changing the description to say that we are using Go 1.13 in unit tests and we are writing the code under Go 1.11 specification or it could work with some older versions.

We have some topics to be discussed:
1. How to define Go version to write a code?
2. How to define Go version used in the CI?
3. When we switch new Go version if a newer version is released?
4. How to guarantee (1) if (1) and (2) are differ?

I don’t have a strong opinion for which Go version should be used. So we rely on the default Go version provided by a GitHub Actions’ environment In this patch. I think it is simple way to say the version to build (tested) and it could work with some older versions if we don’t stick to a specific Go version.

I welcome any comments.

nhoriguchi: Thank you for elaborating the points. Let me share some ideas.

> However, I think it is difficult to guarantee that while updating the code day by day without a CI.

Yes, I agree. Backward compatibility is somehow considered but unfortunately might not perfect. I think that many OSS products are developing implicitly with expecting some kind of compatibility according to the fact that they often declare their requirements like "version x.y.z or later".

> how about changing the description to say that we are using Go 1.13 in unit tests and we are writing the code under Go 1.11 specification or it could work with some older versions.

Yes, that sounds kinder. Or if we like to clarify that the well-tested version is Go 1.13, one option is to mark this version as "recommended" one. I personally think that versions before 1.11 are not recommendable now because of Go module, which has some impact on installation process and might reduce compatibility.

> 1. How to define Go version to write a code?

I think that that's determined based on supported version range. Consider that Go 1.14 is released, then what we do is like below:
- Test with Go 1.14 manually or check compatibility with release note. (call this "evaluation task" for later reference)
- If we can't find any critical problem or concern on Go 1.14, then let's keep the supported version range as-is ("1.11 or later").
- If we find some problem or concern on Go 1.14, then let's specify the supported version range like "1.11 to 1.13".
- When the problem/concern on Go 1.14 is solved, then let's change the supported version range to "1.11 or later"
- If someone find a problem on older version (say 1.11) and we finally judged that we won't fix it, then let's change the supported version range to "1.12 or later."

> 2. How to define Go version used in the CI?

The latest version in supported version range seems preferable to me.

> 3. When we switch new Go version if a newer version is released?

That depends on when the "evaluation task" is done. When to do the task depends on assignee's schedule.

sergefdrv: Can we specify an exact version of Go to be used in GitHub Actions? 

In any case, I would not worry too much about Go version at this stage. The important thing is that the readme gives a good hint of which version will make it working. Maybe we could mention that the lowest recommended version is 1.11, and also refer to the actual version used in CI.

.github/workflows/main.yml
```diff
@@ -8,7 +8,7 @@ env:
 
 jobs:
   build:
-    name: Continuous integration
+    name: Build
```
sergefdrv: How about "CI"? It would be a bit more precise.

ynamiki: I found that we should put a build configuration name like "Build with Go 1.11", etc. here. I put "Continuous integration" as a name of a workflow, which is upper layer of a job.

README.md
ynamiki: > In any case, I would not worry too much about Go version at this stage.

I agree.

I try to build with both Go 1.11 (lowest recommended version) and 1.13 (latest version) in the CI. Both versions are explicitly specified. But I don't specify a patch version so the latest patch version (1.11.13 and 1.13.5 for now) will be used. I think it is trivial matter for us.

ynamiki:
I have updated the patch; now the CI build with both Go 1.11 (lowest recommended version) and 1.13 (latest version). README has been updated to state that our project supports Go 1.11 or later.

ynamiki:
When a new version of Go (1.14?) is released in the future, I will update the workflow configuration to replace 1.13 to 1.14 manually.

sergefdrv:
This PR looks good. It just needs to be rebased on the latest `master` branch.

sergefdrv (approved):
There was a known sporadic test failure that is going to be fix with #110. So I just restarted the CI jobs.

nhoriguchi (approved):
Thanks!

PR #145, opened 2019-12-19
core/internal/requestlist: Fix test case parsing

sergefdrv:
This pull request fixes unit test of `core/internal/requestlist` package.

### DIFF
--- a/core/internal/requestlist/request-list_test.go
+++ b/core/internal/requestlist/request-list_test.go
@@ -31,11 +31,11 @@ var messageImpl = protobufMessages.NewImpl()
 
 func TestList(t *testing.T) {
 	var cases []struct {
-		cid  int // clientID
-		seq  int
-		add  bool
-		rm   bool
-		list map[int]int // clientID -> seq
+		Cid  int // clientID
+		Seq  int
+		Add  bool
+		Rm   bool
+		List map[int]int // clientID -> seq
 	}
 	casesYAML := []byte(`
 - {                           list: {          }}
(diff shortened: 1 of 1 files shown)


nhoriguchi (commented):
I'm not sure why you have to extend visibility of fields in internal structure here. Is that to fix any specific issue on testing? I guess that this is a preparatory work for coming changes, if so, that's OK.

ynamiki (commented):
Exactly the same comment from Naoya.

sergefdrv:
> I'm not sure why you have to extend visibility of fields in internal structure here. Is that to fix any specific issue on testing?

Sorry, I should have clarified that in the commit message. I noticed this when trying to rework unit test of `viewstate` package in similar manner. The issue is that `yaml.Unmarshal` cannot set unexported fields. Moreover, it silently ignores any field it cannot match against the input, so the fields are left unchanged. That is why I also changed to use `yaml.UnmarshalStrict` which returns error in such case.

Now this commit has to be rebased, I will mention the rationale in the commit message.

sergefdrv:
Summary of [changes](https://github.com/hyperledger-labs/minbft/compare/efcf70ca873c574981cbab09bbc2f7e9eacaf335..c47bfc953e082f8709a445d4d595855cc3e1ab4c): rebase with conflict resolution

nhoriguchi (approved):
Very informative. Thank you for the fix.

ynamiki (approved):
Thanks for the fix!

References:
- [Minimum test code](https://play.golang.org/p/hlPraHlgf1H)
- [`Unmarshal()` document](https://godoc.org/gopkg.in/yaml.v2#Unmarshal)

PR #146, opened 2019-12-23
Fix the build status badge in the README

ynamiki:
The build status badge in the README is not displayed. This patch fixes the image source to show the badge.

According to the [document](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/configuring-a-workflow#adding-a-workflow-status-badge-to-your-repository), it should be displayed without this patch but I cannot identify a cause. This patch uses another way (not by a file path but by a name) to specify the URL of the badge.


nhoriguchi (approved):
Screenshot from 2019-12-23 14-03-28

Looks good to me, thank you :+1:

sergefdrv (approved):
Thanks for identifying the issue and finding the solution 👍

PR #147, opened 2019-12-30
Redesign viewstate package

### LINKED ISSUE #110: Fix viewstate package

sergefdrv:
This pull request significantly changes interface and implementation of `core/internal/viewstate` package. Taking into advantage some conclusions about protocol invariants and requirements, the component's interface, usage, and implementation are greatly simplified. For more details, see individual commit messages.

As a side-effect, this pulls request also resolves #110.


core/internal/viewstate/view-state.go
```diff
@@ -17,276 +17,89 @@
...
+// i.e. the current view has become active.
 type State interface {
-	WaitAndHoldActiveView() (view uint64, release func())
-	WaitAndHoldView(view uint64) (ok bool, release func())
-	RequestViewChange(newView uint64) bool
-	StartViewChange(newView uint64) (ok bool, release func())
-	FinishViewChange(newView uint64) (ok bool, release func())
+	HoldView() (current, expected uint64, release func())
+	AdvanceExpectedView(view uint64) (ok bool, release func())
+	AdvanceCurrentView(view uint64) (ok, active bool, release func())
```
nhoriguchi, line 53: The first return value means that expected view or current view is advanced or not, so renaming `ok` with `advanced` looks more self-descriptive to me.

core/internal/viewstate/view-state.go
```diff
@@ -17,276 +17,89 @@
...
-
-	// Last active view number
-	lastActiveView uint64
-
-	// Last requested view change
-	lastRequestedView uint64
-
-	// Last started view change
-	lastNewView uint64
+	sync.RWMutex
```
nhoriguchi, line 57: Introducing `RWMutex` here looks very effective to me, because accesses to current/expected view are mostly read accesses.

core/request.go
```diff
@@ -166,19 +168,23 @@ func makeRequestProcessor(captureSeq requestSeqCapturer, pendingReq requestlist.
 
 		pendingReq.Add(request)
 
-		if err := applyRequest(request); err != nil {
+		currentView, expectedView, releaseView := viewState.HoldView()
+		defer releaseView()
+
+		if currentView != expectedView {
+			return true, nil
```
nhoriguchi, line 175: Returning `true` for `new` means that MessageStreamHandler considers that the request is properly handled (leaving "Handled %s" log), but actually in this case the request seems not to be processed. So could that cause confusion?

core/message-handling.go
```diff
@@ -417,29 +417,38 @@ func makeUIMessageProcessor(captureUI uiCapturer, processViewMessage viewMessage
...
+			defer release()
+
+			if currentView == expectedView {
+				active = true
+			}
+
+			if messageView < currentView {
 				return false, nil
+			} else if messageView > currentView {
+				return false, fmt.Errorf("Message refers to unexpected view")
```
nhoriguchi, line 449: It's a bit unclear to me why `messageView > currentView` case is considered as error and `messageView < currentView` case is not. Maybe that will get clearer when more view-change patches are submitted later, but it's helpful to leave some comment for background of the choice.

nhoriguchi (commented):
Thanks for moving forward to implement view-change. I left some comments/questions but mainly I do agree with whole ideas on this proposal.

core/internal/viewstate/view-state.go
sergefdrv: The idea behind the name `ok` was to indicate if the requested operation - advancing the view number - was successfully performed by the invocation. Renaming it to `advanced` would indeed make it even more straightforward.

core/request.go
sergefdrv: I think this may be a grey area for that return value. On one hand, the request identifier is captured and the request is added to the pending request set before checking this. On the other hand, the replica has not started timers, nor generated a Prepare message at this point. However, the request identifier was captured as new, and the message was processed as much as the replica could do in its current state...

core/message-handling.go
sergefdrv: Good point. If the message refers to an old view then it is simply considered as delayed due to asynchrony in communication; so it gets ignored. However, if the message refers to a future view then the sender is considered faulty, and the error is reported. The latter condition is based on two arguments. 

First, any correct replica, when it starts participating in a new view, must make sure that all other correct replicas will do the same, as described in commit message of "core: Do not wait for active view". 

Second, we use reliable FIFO channels for communication and handle messages received from the same channel sequentially. This implies that messages are processed in the same order as they are sent by the peer. 

Thus, putting it all together, when a replica processes a normal-case message from a correct replica, it must have already transitioned into the corresponding view (or a later view), because the sender must have ensured that before sending the message.

sergefdrv: I am thinking of adding a comment before returning the error, like: "Correct peer replicas ensure that other replicas transition into a new view before sending them any normal-case message for that view."

nhoriguchi: > First, any correct replica, when it starts participating in a new view, must make sure that all other correct replicas will do the same, as described in commit message of "core: Do not wait for active view".
> 
> Second, we use reliable FIFO channels for communication and handle messages received from the same channel sequentially. This implies that messages are processed in the same order as they are sent by the peer.

I've overlooked that message order is guaranteed to be preserved by gRPC layer. I agree that the 2 points explain the reason well. Thank you.

> I am thinking of adding a comment before returning the error, like: "Correct peer replicas ensure that other replicas transition into a new view before sending them any normal-case message for that view."

Sounds nice.

core/request.go
nhoriguchi: We expect that we will restart to process the request in pending request set after view change completes in some mechanism (not implemented yet), so introducing another branch to return silently might be an option.

core/internal/viewstate/view-state.go
ynamiki: I'm fine with both `ok` and `advanced`. Using `ok` for the meaning is natural for me and I have seen many usages.

core/internal/viewstate/view-state.go
```diff
@@ -101,21 +72,14 @@ type State interface {
...
-// In the normal mode, if lastNewView == lastActiveView. Otherwise, in
-// the view-change mode, and lastNewView > lastActiveView.
 type viewState struct {
 	sync.Mutex
 
-	// Last active view number
-	lastActiveView uint64
+	currentView   uint64
+	expectedView  uint64
+	requestedView uint64
```
ynamiki: Even we increment a view 1 million times per second, 585 thousand years are needed to consume `uint64` value. So we don't need to consider exhausting these values now (and also in the future).

ynamiki (approved):
Thanks for the improvement. I left some comments, but they are not requests for changes.

core/internal/viewstate/view-state.go
sergefdrv: Then I'd stick to `ok`. Thanks,

core/message-handling.go
sergefdrv: > I've overlooked that message order is guaranteed to be preserved by gRPC layer.

This is also an [API requirement](https://godoc.org/github.com/hyperledger-labs/minbft/api#ReplicaConnector): "The connection guarantees authenticated destination, eventual delivery, preserving integrity and the original order of messages."

core/request.go
sergefdrv: We could refine diagnostic message logging later, before making the first release. So far, we cannot return `new == false` here because the message is new and not dropped. On the other hand, returning `new == true` causes `Handled ...` message with _debug_ level.

core/internal/viewstate/view-state.go
sergefdrv: Good observation :+1:

sergefdrv:
Summary of [changes](https://github.com/hyperledger-labs/minbft/compare/dbf8db0460ba2ca17a2acf30a4347a2044f6e139..31ab4a9d9a8db1d39e817f1fb95f68917f4b78a3): 

- Added comment to explain "Message refers to unexpected view" error.

ynamiki (approved):

nhoriguchi (approved):

PR #148, opened 2020-01-03
Restructure Protobuf-based message implementation

### LINKED ISSUE #6: Include message type to signed payload
The message type should be covered by signature/USIG certificate to guarantee that a singed payload will not happen to be misinterpreted. NB: Protobuf does not include the message type into the binary encoding.

sergefdrv:
This pull request restructures Protobuf-based message implementation in order to solve its two main problems: non-determinism and possible misinterpretation of messages' authenticated payload. Check individual commit messages for more details.

Resolves #6


messages/protobuf/impl.go
```diff
@@ -32,7 +31,7 @@ func NewImpl() messages.MessageImpl {
 func (*impl) NewFromBinary(data []byte) (messages.Message, error) {
 	msg := &pb.Message{}
 	if err := proto.Unmarshal(data, msg); err != nil {
-		return nil, fmt.Errorf("failed to unmarshal message wrapper: %s", err)
+		return nil, errors.Wrap(err, "failed to unmarshal message wrapper")
```
ynamiki: As you may know, Go 1.13 introduces [new build-in error handling features](https://blog.golang.org/go1.13-errors) and there is a compatibility package, [golang.org/x/xerrors](https://github.com/golang/go/wiki/ErrorValueFAQ) for earlier versions. How about using the new feature?

To use the feature, just replace `%s` to `%w`:
```
fmt.Errorf("failed to unmarshal message wrapper: %w", err)
```

sergefdrv: Thank you for pointing this out; I actually overlooked this new feature. The new feature looks almost the same as what `pkg/errors` provide. One notable difference is that `pkg/errors` automatically annotates errors with stack trace, which might be useful when analyzing problems and debugging. What do you think?

sergefdrv: I checked more carefully and found that `xerrors` package in fact annotates errors with stack trace, as well. Then I think we should rather use it in favor of `pkg/error` package. 

The only question if we should give up on supporting Go 1.11 or use the compatibility package? Given that this is a component to be used by other projects, I tend to think of preserving better compatibility, even though it is less convenient.

ynamiki: `golang.org/x/xerrors` supports Go 1.11 or later:
https://github.com/golang/xerrors/blob/master/go.mod

So we don't need to update our compatibility policy.

sergefdrv: Should we also include such a `go` directive in our `go.mod`?

sergefdrv:
Summary of [changes](https://github.com/hyperledger-labs/minbft/compare/97045dbc76355063c33020557fc1d7f73146a10c..e9754f08ba494605d3207caa414b0ebb12dee72a):

- Replaced `pkg/errors` with `xerrors`

messages/protobuf/impl.go
ynamiki: Yes, I will do. Thanks for creating the issue.

ynamiki (approved):
Thanks for update. Looks good to me.

nhoriguchi (approved):
Sorry for late response, the suggested change looks nice.

The non-determinism of protobuf's serialization was new to me.  As described in [godoc](https://godoc.org/github.com/golang/protobuf/proto#Buffer.SetDeterministic) `deterministic` flag doesn't guarantee stable serialization over time, so it's recommended to "define their own canonicalization specification and implement their own serializer rather than relying on this API." That's what `authen.go` does.

PR #153, opened 2020-01-13
Ignore errors when processing valid embedded messages

sergefdrv:
This pull request will allow to use the existing code to process messages embedded into ViewChange and NewView messages, i.e. messages of the message log included into ViewChange message, and ViewChange messages of the new-view certificate included into NewView message. Please check the commit message for more details.

### DIFF
--- a/core/message-handling_test.go
+++ b/core/message-handling_test.go
@@ -256,7 +256,7 @@ func TestMakeReplicaMessageProcessor(t *testing.T) {
 		args := mock.MethodCalled("uiMessageProcessor", msg)
 		return args.Bool(0), args.Error(1)
 	}
-	process := makeReplicaMessageProcessor(id, processMessage, processUIMessage)
+	process := makeReplicaMessageProcessor(id, processMessage, processUIMessage, logging.MustGetLogger(module))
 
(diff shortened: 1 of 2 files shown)


core/message-handling_test.go
```diff
@@ -275,8 +275,9 @@ func TestMakeReplicaMessageProcessor(t *testing.T) {
 		prepare := messageImpl.NewPrepare(primary, viewForPrimary(n, primary), request)
 
 		mock.On("messageProcessor", request).Return(false, fmt.Errorf("Error")).Once()
+		mock.On("uiMessageProcessor", prepare).Return(true, nil).Once()
 		_, err = process(prepare)
-		assert.Error(t, err, "Failed to process embedded Request")
+		assert.NoError(t, err)
```
ynamiki, line 280: It is my understanding that we will get a warning log at this line but it doesn't matter whether the log is generated or not (i.e. there is an ignored message or not) in this test case.

ynamiki (approved):

nhoriguchi (approved):

core/message-handling_test.go
sergefdrv: Exactly, this test does only check protocol logic and ignores diagnostic output.

PR #165, opened 2020-01-30
Introduce environment checker utility

### LINKED ISSUE #150: Environment checker in build process
We have non-trivial prerequisite steps before running [Running Example](https://github.com/hyperledger-labs/minbft#running-example), which can be a barrier for early users. So some environment check and guidance messages might be helpful. For example, if we try to build on non-SGX system with SGX_MODE=HW, the warning like below should be helpful:
~~~
$ make install
Platform doesn't support Intel SGX. Please retry with environment variable SGX_MODE=SIM.
~~~

Similarly, if we try to build without meeting requirement, the error like below works well.
~~~
$ make install
Intel SGX SDK is not available on current platform, see README for requirements.
~~~

nhoriguchi:
This pull request is suggesting to introduce a tool to check whether required libraries and settings are properly added. According to [the document](https://software.intel.com/en-us/articles/properly-detecting-intel-software-guard-extensions-in-your-applications), we can use `sgx_cap_get_status()` to detect SGX availability, so this tool does it. I expect that it would help users find setting issues for their own.

I show below how `make prerequisite-check` works as examples:
- when the system supports and enables SGX:
  ~~~
  ubuntu@sgx:~/minbft$ make prerequisite-check
  OS: Ubuntu 18.04.3 LTS
  Go version: 1.13.6
  Intel SGX SDK version: 2.8.100.3
  Intel SGX is supported and enabled on this system.
  Intel SGX PSW version: 2.8.100.3-bionic1
  Enviroment check passed.
  ~~~
- when SGX is unavailable:
  ~~~
  ubuntu@nosgx:~/minbft$ make prerequisite-check
  OS: Ubuntu 18.04.3 LTS
  Go version: 1.13.6
  Intel SGX SDK version: 2.3.101.46683
  This CPU does not support Intel SGX.
  Makefile:8: recipe for target 'check' failed
  make[1]: *** [check] Error 1
  Intel SGX is not available. Please set environment variable SGX_MODE=SIM.
  Makefile:82: recipe for target 'prerequisite-check' failed
  make: *** [prerequisite-check] Error 1
  ~~~
- when SGX is unavailable and `SGX_MODE=SIM` is set:
  ~~~
  ubuntu@nosgx:~/minbft$ SGX_MODE=SIM make prerequisite-check
  OS: Ubuntu 18.04.3 LTS
  Go version: 1.13.6
  Intel SGX SDK version: 2.3.101.46683
  Enviroment check passed.
  ~~~

Expected to resolve #150.

Some notes which we might want to discuss:
- I mainly borrowed code from [`sgx_enable.c`](https://github.com/intel/sgx-software-enable/blob/master/sgx_enable.c) for `check-sgx-capability.c`, but substantially removed lines because it does "software-enable" part not only detecting SGX support. In this case, how do we handle license matter?
- Currently `prerequisite-check.sh` aborts if version check fails, but that might be too strict. If you think just warning is fine, I'll do it. And the script also has an assumption of using Ubuntu, that might be undesirable limitation in a long run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ help:
 	@echo '  usig-*           - Make USIG target, where target is one of:'
 	@echo '                     $(usig-target-list)'
 
-build: usig-build
+build: prerequisite-check usig-build
 	go build -o $(builddir)/keytool ./sample/authentication/keytool
 	go build -o $(builddir)/peer ./sample/peer
 
--- a/Makefile
+++ b/Makefile
@@ -78,5 +78,8 @@ lint:
 generate:
 	go generate ./...
 
+prerequisite-check:
+	@bash tools/prerequisite-check.sh
+
 $(usig-target-list):
 	$(MAKE) -C usig/sgx $(patsubst usig-%,%,$@)
--- a/tools/prerequisite-check.sh
+++ b/tools/prerequisite-check.sh
@@ -1,0 +1,131 @@
+#!/bin/bash
+
+REQUIRED_GO_VERSION=1.11
+REQUIRED_SGXSDK_VERSION=2.3.101
+
+show_os_info() {
+	echo "OS: $(. /etc/os-release ; echo $PRETTY_NAME)"
+}
+
+compare_version() {
+	local ver1=$(echo "$1" | awk -F. '{ printf("%d%03d%03d%05d\n", $1,$2,$3,$4); }';)
+	local ver2=$(echo "$2" | awk -F. '{ printf("%d%03d%03d%05d\n", $1,$2,$3,$4); }';)
+
+	if [ "$ver1" -lt "$ver2" ] ; then
+		return 1
+	else
+		return 0
+	fi
+}
+
+check_go_version() {
+	if [[ "$(go version)" =~ go([0-9]+\.[0-9]+\.?[0-9]*) ]] ; then
+		local goversion=${BASH_REMATCH[1]}
+	else
+		echo "Failed to parse go version. Please install it."
+		return 1
+	fi
+
+	if ! compare_version "$goversion" "$REQUIRED_GO_VERSION" ; then
+		echo "Your Go (version $goversion) is older than required version ($REQUIRED_GO_VERSION), so please update it."
+		return 1
+	fi
+
+	echo "Go version: $goversion"
+}
+
+check_sgxsdk_version() {
+	local sgxsdk_version=$(grep ^Version: $SGX_SDK/pkgconfig/libsgx_urts.pc | awk '{print $2}')
+	if [ ! "$sgxsdk_version" ] ; then
+		echo "Failed to parse SGX SDK version from $SGX_SDK/pkgconfig/libsgx_urts.pc. Please make sure that SGX SDK is properly installed."
+		return 1
+	fi
+
+	if ! compare_version $sgxsdk_version $REQUIRED_SGXSDK_VERSION ; then
+		echo "Your SGX SDK (version $sgxsdk_version) is older than required version ($REQUIRED_SGXSDK_VERSION), so please update it."
+		return 1
+	fi
+
+	echo "Intel SGX SDK version: $sgxsdk_version"
+}
+
+check_sgx_support() {
+	if [ "$SGX_MODE" == SIM ] ; then
+		return 0
+	fi
+
+	if make -s --no-print-directory -C tools/sgx-capability check ; then
+		return 0
+	else
+		echo "Intel SGX is not available. Please set environment variable SGX_MODE=SIM."
+		return 1
+	fi
+}
+
+get_sgx_psw_version() {
+	local distro="$(. /etc/os-release ; echo $PRETTY_NAME)"
+	local version=
+
+	if [[ "$distro" =~ "Ubuntu" ]] ; then
+		version=$(dpkg -s libsgx-enclave-common 2> /dev/null | grep '^Version:' | awk '{print $2}')
+	elif [[ "$distro" =~ "Red Hat Enterprise Linux" ]] ; then
+		version=$(rpm -qi libsgx-enclave-common 2> /dev/null | grep '^Version' | awk '{print $3}')
+	elif [[ "$distro" =~ "CentOS" ]] ; then
+		version=$(rpm -qi libsgx-enclave-common 2> /dev/null | grep '^Version' | awk '{print $3}')
+	else
+		echo "Your Linux distribution '$os' is not supported now." >&2
+		return 1
+	fi
+
+	if [ "$version" ] ; then
+		echo "$version"
+		return 0
+	else
+		echo "Failed to get SGX PSW version." >&2
+		return 1
+	fi
+}
+
+check_sgxpsw_version() {
+	if [ "$SGX_MODE" == SIM ] ; then
+		return 0
+	fi
+
+	local version="$(get_sgx_psw_version)"
+	if [ "$version" ] ; then
+		echo "Intel SGX PSW version: $version"
+		return 0
+	else
+		echo "Intel SGX PSW is not installed in your system. Please install it or set environment variable SGX_MODE=SIM."
+		return 1
+	fi
+}
+
+show_os_info
+
+if ! check_go_version ; then
+	exit 1
+fi
+
+if [ ! "$SGX_SDK" ] ; then
+	echo "Environment variable SGX_SDK not set, please make sure to source the environment file given by SGX SDK."
+	exit 1
+fi
+
+if [ "$SGX_MODE" == SIM ] ; then
+	echo "SGX is running in simulation mode."
+fi
+
+if ! check_sgxsdk_version ; then
+	exit 1
+fi
+
+if ! check_sgx_support ; then
+	exit 1
+fi
+
+if ! check_sgxpsw_version ; then
+	exit 1
+fi
+
+echo "Enviroment check passed."
--- a/tools/prerequisite-check.sh
+++ b/tools/prerequisite-check.sh
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+if [ "$SKIP_PREREQUISITE_CHECK" ] ; then
+	echo "Skip prerequisite_check."
+	exit 0
+fi
+
 REQUIRED_GO_VERSION=1.11
 REQUIRED_SGXSDK_VERSION=2.3.101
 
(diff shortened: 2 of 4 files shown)


ynamiki:
> I mainly borrowed code from sgx_enable.c for check-sgx-capability.c, but substantially removed lines because it does "software-enable" part not only detecting SGX support. In this case, how do we handle license matter?

Original code is [licensed](https://github.com/intel/sgx-software-enable#license) under [BSD 3-Clause](https://opensource.org/licenses/BSD-3-Clause). It is [said](https://softwareengineering.stackexchange.com/questions/40561/is-bsd-license-compatible-with-apache) that BSD 3-Clause is compatible with Apache 2.0 (our license). I think what we should do is only copying (keeping) the comments about copyright and license on the top of the original file (this is requirement stated in the license). That is, you have modified a BSD-licensed file and incorporated it into our Apache-licensed project.

Note that, I am not a lawyer.

nhoriguchi:
> > I mainly borrowed code from sgx_enable.c for check-sgx-capability.c, but substantially removed lines because it does "software-enable" part not only detecting SGX support. In this case, how do we handle license matter?
> 
> Original code is [licensed](https://github.com/intel/sgx-software-enable#license) under [BSD 3-Clause](https://opensource.org/licenses/BSD-3-Clause). It is [said](https://softwareengineering.stackexchange.com/questions/40561/is-bsd-license-compatible-with-apache) that BSD 3-Clause is compatible with Apache 2.0 (our license). I think what we should do is only copying (keeping) the comments about copyright and license on the top of the original file (this is requirement stated in the license). That is, you have modified a BSD-licensed file and incorporated it into our Apache-licensed project.

I updated the license header of the file, thank you!

tools/sgx-capability/check-sgx-capability.c
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * Copyright (C) 2011-2019 Intel Corporation. All rights reserved.
```
sergefdrv, line 21: Maybe also mention where the BSD 3-Clause code comes from?

tools/sgx-capability/check-sgx-capability.c
```diff
@@ -0,0 +1,106 @@
...
+		return 0;
+	} else if (result == SGX_DISABLED_UNSUPPORTED_CPU) {
+		printf("This CPU does not support Intel SGX.\n");
+		return 1;
+	} else if (result == SGX_DISABLED_LEGACY_OS) {
+		printf("This processor supports Intel SGX but was booted in legacy mode.\n");
+		printf("A UEFI boot is required to determine whether or not your BIOS\n");
+		printf("supports Intel SGX, and to attempt the software enable.\n");
+		printf("Alternatively, you can check your BIOS to see if it has an\n");
+		printf("explicit option to enable SGX.\n");
```
sergefdrv: Maybe we should not mention software enable in the output as it might be a bit confusing? I suggest the following output text:

```
This processor supports Intel SGX but was booted in legacy mode.
A UEFI boot is required to determine whether or not your BIOS
supports Intel SGX.
```

tools/prerequisite-check.sh
```diff
@@ -0,0 +1,104 @@
...
+
+if ! check_go_version ; then
+	exit 1
+fi
+
+if [ ! "$SGX_SDK" ] ; then
+	echo "Environment variable SGX_SDK not set, please make sure to source the environment file given by SGX SDK."
+	exit 1
+fi
+
```
sergefdrv, line 123: Maybe output something here if SGX simulation mode is used?

sergefdrv (approved):
Thank you, I think this is a great improvement in building experience.

tools/prerequisite-check.sh
```diff
@@ -0,0 +1,104 @@
...
+		return 1
+	fi
+}
+
+check_sgx_psw_installed() {
+	if [ "$SGX_MODE" == SIM ] ; then
+		return 0
+	fi
+
+	dpkg -l libsgx-enclave-common > /dev/null
```
ynamiki: I think it is too strict about the assumption of Ubuntu, because Intel SGX [officially supports](https://01.org/intel-softwareguard-extensions/downloads/intel-sgx-linux-2.3.1-release) not only Ubuntu but also Red Hat Enterprise Linux, CentOS, Fedora, and SUSE Linux Enterprise Server, and our code don't depends on specific operating system.

How about checking existence of `/opt/intel/sgxpsw/uninstall.sh` here not to depend on OS-specific package managers? According to the document ([v2.3.1](https://download.01.org/intel-sgx/linux-2.3.1/docs/Intel_SGX_Installation_Guide_Linux_2.3.1_Open_Source.pdf) and [v2.8](https://download.01.org/intel-sgx/sgx-linux/2.8/docs/Intel_SGX_Installation_Guide_Linux_2.8_Open_Source.pdf) (the latest release), p. 12), Intel SGX PSW installs the script in the path.

tools/sgx-capability/check-sgx-capability.c
nhoriguchi: Ok, I'll mention it.

tools/sgx-capability/check-sgx-capability.c
nhoriguchi: Thanks, I'll update this message.

tools/prerequisite-check.sh
nhoriguchi: Sounds helpful. I'll add it.

tools/prerequisite-check.sh
nhoriguchi: > How about checking existence of `/opt/intel/sgxpsw/uninstall.sh` here not to depend on OS-specific package managers? According to the document ([v2.3.1](https://download.01.org/intel-sgx/linux-2.3.1/docs/Intel_SGX_Installation_Guide_Linux_2.3.1_Open_Source.pdf) and [v2.8](https://download.01.org/intel-sgx/sgx-linux/2.8/docs/Intel_SGX_Installation_Guide_Linux_2.8_Open_Source.pdf) (the latest release), p. 12), Intel SGX PSW installs the script in the path.

It seems that `/opt/intel/sgxpsw/uninstall.sh` is not installed when installing SGX PSW from Debian repo, so we can't rely on this file as generic solution. Actually, the installation guide mentions the installer `sgx_linux_<os>_x64_psw_<version>.bin` in "Uninstall the Intel®SGX PSW package" section, but that file does not exist, so this paragraph might be outdated.

I thought that we could use `/usr/lib/x86_64-linux-gnu/libsgx_enclave_common.so.1` as a indicator, but that's located differently depending on Linux distributions (for example it's `/usr/lib64/libsgx_enclave_common.so.1` for RHEL). Another thought is to match "libsgx_enclave_common.so" over `find /usr/` or `ldconfig -p` output, although it might look like a hack.

ynamiki: Thanks for the info, I have no good idea and agree with this should not be complicated process; it is no problem with proceeding with this implementation.

Makefile
```diff
@@ -78,5 +78,8 @@ lint:
 generate:
 	go generate ./...
 
+prerequisite-check:
+	@bash tools/prerequisite-check.sh
```
ynamiki, line 82: Can we add a path to skip this check like `make PREREQUISITE_CHECK=0` or another way? It is useful for an expert who wants to run this program not on a recommended environment.

nhoriguchi: Sure, thanks for the suggestion. I'll define `SKIP_PREREQUISITE_CHECK` to do prerequisite check by default.

nhoriguchi:
Summary of changes:
- improved output messages.
- `check-sgx-capability.c`: added reference to original source file,
- `prerequisite-check.sh`: renamed `check_sgx_psw_installed` with `check_sgxpsw_version` (for consistency with `check_sgxsdk_version`)
- introduced `get_sgx_psw_version` (with adding some support for RHEL/CentOS, but we still have no generic solution,)

tools/prerequisite-check.sh
sergefdrv: What configure scripts usually do to detect missing dependencies is try building a tiny program (usually inlined in the script) and check if the dependency is resolved. But this may be too complicated for us at this stage.

tools/prerequisite-check.sh
```diff
@@ -0,0 +1,131 @@
...
+if ! check_go_version ; then
+	exit 1
+fi
+
+if [ ! "$SGX_SDK" ] ; then
+	echo "Environment variable SGX_SDK not set, please make sure to source the environment file given by SGX SDK."
+	exit 1
+fi
+
+if [ "$SGX_MODE" == SIM ] ; then
```
sergefdrv:
```suggestion
if [ "$SGX_MODE" != HW ] ; then
```

cf. https://github.com/hyperledger-labs/minbft/blob/335b81417ed63a3ebc2ef9805407fc58e8afeb98/usig/sgx/Makefile#L103.

sergefdrv (changes requested):
Looks very good, just a small comment.

tools/prerequisite-check.sh
sergefdrv: Sorry that I missed that in the last review round, but there are a couple more such comparisons. Namely, in `check_sgx_support` and `check_sgxpsw_version`.

sergefdrv: We could do the other way around and always build in hardware mode unless `"$SGX_MODE" == "SIM"`. But this would not match the convention given by SGX SDK...

ynamiki (approved):

tools/prerequisite-check.sh
nhoriguchi: Thanks for the input. I found that if we use `"$SGX_MODE" != "HW"`, default case (leaving `$SGX_MODE` empty) is considered as simulation mode, which seems inconsistent with `usig/sgx/Makefile` (which sets `SGX_MODE ?= HW`, so running in hardware mode by default). So I'll fix it by giving default value of `SGX_MODE` inside the script.

> We could do the other way around and always build in hardware mode unless `"$SGX_MODE" == "SIM"`. But this would not match the convention given by SGX SDK...

I'm OK to use simulation mode If `SGX_MODE` is set to some string other than "", "HW" and "SIM". The check script will show the running mode in such a case.

sergefdrv (approved):

PR #166, opened 2020-03-06
forward REQUEST messages to primary when prepare timer expires

### LINKED ISSUE #5: Consider faulty client

### LINKED ISSUE #140: Relay request messages to primary replica
As discussed in #127, we need to implement forwarding request messages to primary replica to completely solve issue #5.

nhoriguchi:
This pull request proposes to introduce request forwarding to handle some faulty case in request submission. This allows the MinBFT cluster to accept a request in some cases when the primary does not receive it from the client. The table below provides brief summary of how the behavior changes:

| case           | origin | patched | 
|----------------|--------|---------| 
| R0R1R2         | pass   | pass    | 
| R0             | fail   | fail    | 
| R0R1           | pass   | pass    | 
| R1             | fail   | fail    | 
| R1R2           | fail   | pass    | 
| R0R1Q2         | pass   | pass    | 
| R0Q1           | pass   | pass    | 
| R1Q2           | fail   | pass    | 

Here we consider a minimum typical cluster (N is 3 and replica 0 is the primary.) Case ID shows which replica(s) a client sends REQUEST to. For example, in case "R0R1" a client sends REQUEST only to replica 0 and 1. In another example "R0R1Q2", a client sends proper REQUEST messages to replica 0 and 1, and a modified REQUEST message (with the same seq) to replica 2. In the result columns, "pass" means that the cluster reaches a consensus.

The most interesting case is "R1R2", where a client sends the proper REQUEST messages to replica 1 and 2. This case becomes "pass" due to request forwarding.

What I noticed while working on this is that case "R1" fails even with this patch. This is because in current implementation the replica could send REPLY to the client only when the replica received the REQUEST directly from the client (i.e. a replica who got a forwarded request can't send REPLY.) Similarly, a backup replica which didn't receive REQUEST from the client and processed REQUEST embedded in the PREPARE message can't send REPLY message to its requester. I guess that these are not due to some specific reason, but it's just implemented now as such. So we can do some to fix them in the future if we think it's worth doing.

Resolves #140
Resolves #5

### DIFF
--- a/core/message-handling.go
+++ b/core/message-handling.go
@@ -45,11 +45,11 @@ type messageStreamHandler func(in <-chan []byte, reply chan<- []byte)
 // processed before.
 type incomingMessageHandler func(msg messages.Message, own bool) (reply <-chan messages.Message, new bool, err error)
 
-// peerMessageSupplier supplies messages for peer replica.
+// messageSupplier supplies messages for peer replica.
 //
 // Given a channel, it supplies the channel with messages to be
 // delivered to the peer replica.
-type peerMessageSupplier func(out chan<- []byte)
+type messageSupplier func(out chan<- []byte)
 
 // peerConnector initiates message exchange with a peer replica.
 //
@@ -143,7 +143,7 @@ type generatedMessageConsumer func(msg messages.ReplicaMessage)
 // defaultIncomingMessageHandler construct a standard
 // incomingMessageHandler using id as the current replica ID and the
 // supplied interfaces.
-func defaultIncomingMessageHandler(id uint32, log messagelog.MessageLog, config api.Configer, stack Stack, logger *logging.Logger) incomingMessageHandler {
+func defaultIncomingMessageHandler(id uint32, log messagelog.MessageLog, config api.Configer, stack Stack, logger *logging.Logger, requestForward map[uint32]messagelog.MessageLog) incomingMessageHandler {
 	n := config.N()
 	f := config.F()
 
@@ -172,7 +172,7 @@ func defaultIncomingMessageHandler(id uint32, log messagelog.MessageLog, config 
 	handleReqTimeout := makeRequestTimeoutHandler(requestViewChange, logger)
 	startReqTimer := makeRequestTimerStarter(clientStates, handleReqTimeout, logger)
 	stopReqTimer := makeRequestTimerStopper(clientStates)
-	startPrepTimer := makePrepareTimerStarter(clientStates, logger)
+	startPrepTimer := makePrepareTimerStarter(n, clientStates, logger, requestForward)
 	stopPrepTimer := makePrepareTimerStopper(clientStates)
 
 	countCommitment := makeCommitmentCounter(f)
@@ -253,14 +253,18 @@ func makeMessageStreamHandler(handle incomingMessageHandler, logger *logging.Log
 
 // startPeerConnections initiates asynchronous message exchange with
 // peer replicas.
-func startPeerConnections(replicaID, n uint32, connector api.ReplicaConnector, log messagelog.MessageLog, logger *logging.Logger) error {
-	supply := makePeerMessageSupplier(log)
-
+func startPeerConnections(replicaID, n uint32, connector api.ReplicaConnector, log messagelog.MessageLog, logger *logging.Logger, requestforward map[uint32]messagelog.MessageLog) error {
 	for peerID := uint32(0); peerID < n; peerID++ {
 		if peerID == replicaID {
 			continue
 		}
 
+		// TODO: we need to handle the situation where messagelog accumulates
+		// many requests indefinitely when the destination replica stops
+		// receiving them.
+		requestforward[peerID] = messagelog.New()
+
+		supply := makeMessageSupplier(log, requestforward[peerID].Stream(nil))
 		connect := makePeerConnector(peerID, connector)
 		if err := startPeerConnection(connect, supply); err != nil {
 			return fmt.Errorf("Cannot connect to replica %d: %s", peerID, err)
@@ -272,7 +276,7 @@ func startPeerConnections(replicaID, n uint32, connector api.ReplicaConnector, l
 
 // startPeerConnection initiates asynchronous message exchange with a
 // peer replica.
-func startPeerConnection(connect peerConnector, supply peerMessageSupplier) error {
+func startPeerConnection(connect peerConnector, supply messageSupplier) error {
 	out := make(chan []byte)
 
 	// So far, reply stream is not used for replica-to-replica
@@ -301,11 +305,19 @@ func handleGeneratedPeerMessages(log messagelog.MessageLog, handle incomingMessa
 	}
 }
 
-// makePeerMessageSupplier construct a peerMessageSupplier using the
+// makeMessageSupplier constructs a messageSupplier using the
 // supplied message log.
-func makePeerMessageSupplier(log messagelog.MessageLog) peerMessageSupplier {
+// Another parameter 'forwardChan' is to accept unicasted messages.
+func makeMessageSupplier(log messagelog.MessageLog, forwardChan <-chan messages.Message) messageSupplier {
 	return func(out chan<- []byte) {
-		for msg := range log.Stream(nil) {
+		logChan := log.Stream(nil)
+		for {
+			var msg messages.Message
+
+			select {
+			case msg = <-logChan:
+			case msg = <-forwardChan:
+			}
 			msgBytes, err := msg.MarshalBinary()
 			if err != nil {
 				panic(err)
--- a/core/internal/messagelog/messagelog.go
+++ b/core/internal/messagelog/messagelog.go
@@ -38,15 +38,15 @@ import (
 // indicates the returned channel should be closed. Nil channel may be
 // passed if there's no need to close the returned channel.
 type MessageLog interface {
-	Append(msg messages.ReplicaMessage)
-	Stream(done <-chan struct{}) <-chan messages.ReplicaMessage
+	Append(msg messages.Message)
+	Stream(done <-chan struct{}) <-chan messages.Message
 }
 
 type messageLog struct {
 	lock sync.RWMutex
 
 	// Messages in order added
-	msgs []messages.ReplicaMessage
+	msgs []messages.Message
 
 	// Buffered channels to notify about new messages
 	newAdded []chan<- struct{}
@@ -57,7 +57,7 @@ func New() MessageLog {
 	return &messageLog{}
 }
 
-func (log *messageLog) Append(msg messages.ReplicaMessage) {
+func (log *messageLog) Append(msg messages.Message) {
 	log.lock.Lock()
 	defer log.lock.Unlock()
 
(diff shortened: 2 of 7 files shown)


core/message-handling.go
```diff
@@ -51,6 +51,10 @@ type incomingMessageHandler func(msg messages.Message, own bool) (reply <-chan m
 // delivered to the peer replica.
 type peerMessageSupplier func(out chan<- []byte)
 
+// forwardedRequestSupplier supplies Request messages to forward to
+// the primary replica.
+type forwardedRequestSupplier func(out chan<- []byte, forwardIn chan messages.Request)
```
ynamiki: Can we change type of `forwardIn` to `<-chan` to increase the type-safety?

nhoriguchi: Sure, I fixed. Thanks!

core/message-handling.go
```diff
@@ -251,28 +255,56 @@ func makeMessageStreamHandler(handle incomingMessageHandler, logger *logging.Log
...
 func startPeerConnections(replicaID, n uint32, connector api.ReplicaConnector, log messagelog.MessageLog, logger *logging.Logger) error {
 	supply := makePeerMessageSupplier(log)
+	forwardOut = make(map[uint32](chan messages.Request))
 
 	for peerID := uint32(0); peerID < n; peerID++ {
 		if peerID == replicaID {
 			continue
 		}
 
+		forwardOut[peerID] = make(chan messages.Request)
```
ynamiki: Now we can also use `<-chan` here, Line 258, and 264.

core/message-handling.go
```diff
@@ -251,28 +255,56 @@ func makeMessageStreamHandler(handle incomingMessageHandler, logger *logging.Log
...
+// prepare timer expires.
+func forwardRequestMessage(peerID uint32, msg messages.Request, logger *logging.Logger) {
+	logger.Debugf("Forward REQUEST to replica %d\n", peerID)
+	forwardOut[peerID] <- msg
+}
+
 // startPeerConnection initiates asynchronous message exchange with a
 // peer replica.
-func startPeerConnection(connect peerConnector, supply peerMessageSupplier) error {
+func startPeerConnection(connect peerConnector, supply peerMessageSupplier, forward forwardedRequestSupplier, forwardIn chan messages.Request) error {
```
ynamiki:
```suggestion
func startPeerConnection(connect peerConnector, supply peerMessageSupplier, forward forwardedRequestSupplier, forwardIn <-chan messages.Request) error {
```

ynamiki (commented):
Following the last update, there can be some more updates on the type-safety.

core/message-handling.go
nhoriguchi: When I use `<-chan` on L258/L264/L271, compiler shows the following error on L302:

> invalid operation: cannot send to receive-only type <-chan messages.Request

https://github.com/Naoya-Horiguchi/minbft/blob/fe982365837a46d6326ac363937778ab10e70e8c/core/message-handling.go#L300-L303

L307 can have `<-` as you suggested, so I added it.

ynamiki: > When I use <-chan on L258/L264/L271, compiler shows the following error on L302

Sorry, I overlooked the send operation in L302; They must be bidirectional.

ynamiki (approved):
Thanks, looks good to me.

core/message-handling.go
```diff
@@ -251,28 +255,56 @@ func makeMessageStreamHandler(handle incomingMessageHandler, logger *logging.Log
 	}
 }
 
+var forwardOut map[uint32](chan messages.Request)
```
sergefdrv: This should not be a global variable since it represents a piece of individual replica state.

core/message-handling.go
```diff
@@ -251,28 +255,56 @@ func makeMessageStreamHandler(handle incomingMessageHandler, logger *logging.Log
...
+func makeForwardedRequestSupplier(peerID uint32) forwardedRequestSupplier {
+	return func(out chan<- []byte, in <-chan messages.Request) {
+		for {
+			msg := <-in
+			msgBytes, err := msg.MarshalBinary()
+			if err != nil {
+				panic(err)
+			}
+			out <- msgBytes
+		}
```
sergefdrv: This loop is almost the same as in `makePeerMessageSupplier`. Maybe we can move the common part to a separate helper function? Or pass the forward channel to `makePeerMessageSupplier` and extend it like follows:

```go
for {
	var msg messages.Message

	select {
	case msg = <-forwardChan:
	case msg = <-logChan:
	}

	msgBytes, err := msg.MarshalBinary()
	if err != nil {
		panic(err)
	}
	out <- msgBytes
}
```

core/message-handling.go
```diff
@@ -251,28 +255,56 @@ func makeMessageStreamHandler(handle incomingMessageHandler, logger *logging.Log
...
+	}
+}
+
+// forwardRequestMessage send Request message to primary replica when
+// prepare timer expires.
+func forwardRequestMessage(peerID uint32, msg messages.Request, logger *logging.Logger) {
+	logger.Debugf("Forward REQUEST to replica %d\n", peerID)
+	forwardOut[peerID] <- msg
```
sergefdrv: I'm afraid this may block the caller for indefinite time if the destination replica is faulty. How do we handle this situation?

sergefdrv (changes requested):

sergefdrv:
> What I noticed while working on this is that case "R1" fails even with this patch. This is because in current implementation the replica could send REPLY to the client only when the replica received the REQUEST directly from the client (i.e. a replica who got a forwarded request can't send REPLY.) Similarly, a backup replica which didn't receive REQUEST from the client and processed REQUEST embedded in the PREPARE message can't send REPLY message to its requester. I guess that these are not due to some specific reason, but it's just implemented now as such. So we can do some to fix them in the future if we think it's worth doing.

This is because replicas use received Request messages to distinguish incoming client connection originated at different clients. We could consider improving this at some point.

core/message-handling.go
nhoriguchi: OK, so I'm thinking of adding a new package under core/internal to manage request forwarding, which seems to me better than adding into existing packages.

core/message-handling.go
nhoriguchi: Yes, we can factorize this code. What was beyond my first choice was that REQUEST does not belong to message category "Peer Message", so if we go in this direction, maybe I'll rename `makePeerMessageSupplier` to simply `makeMessageSupplier`.

core/message-handling.go
nhoriguchi: I didn't consider such situation. I think that this code does not block as long as `forwardOut` channel is listened by the goroutine running forwardedRequestSupplier, which is satisfied if the replica is not faulty. How does the destination replica affect this code?

nhoriguchi:
Change summary:

- generalized peer message supplier to handle both log channel and forward channel,
- use `messageLog` instances to handle unicast, which indirectly solves the concern about blocking channel operations because `supplyMessages()` already addresses it.

sergefdrv:
@Naoya-Horiguchi Thank you for doing this, but I'm afraid I barely have time to review it until the beginning of April. On the other hand, I'd still like to vet on this issue and give my feedback. Would it be fine for you to wait for my feedback for a couple more weeks?

nhoriguchi:
> @Naoya-Horiguchi Thank you for doing this, but I'm afraid I barely have time to review it until the beginning of April. On the other hand, I'd still like to vet on this issue and give my feedback. Would it be fine for you to wait for my feedback for a couple more weeks?

Sure, this issue is not urgent, so we can wait until April.

core/message-handling.go
```diff
@@ -45,11 +45,12 @@ type messageStreamHandler func(in <-chan []byte, reply chan<- []byte)
...
 type incomingMessageHandler func(msg messages.Message, own bool) (reply <-chan messages.Message, new bool, err error)
 
-// peerMessageSupplier supplies messages for peer replica.
+// messageSupplier supplies messages for peer replica.
 //
 // Given a channel, it supplies the channel with messages to be
 // delivered to the peer replica.
-type peerMessageSupplier func(out chan<- []byte)
+// Another parameter 'forwardChan' is to accept unicasted messages.
+type messageSupplier func(out chan<- []byte, forwardChan <-chan messages.Message)
```
sergefdrv: I suggest passing `forwardChan` to `makeMessageSupplier` instead. This way would better align with what we already do with the common message log.

sergefdrv (commented):
I left a minor comment, but apart from that, I think this PR is a solid step towards fully solving the request forwarding problem.

Now, using `messagelog` for buffering the forwarded requests eliminates dandling goroutines when the destination replica stops receiving the messages. However, this will still accumulate the forwarded messages in the unicast message logs, whereas we only need to keep at most one request per client (until we add support for concurrent client requests). I think we can try to improve this as a separate issue.

nhoriguchi:
I've updated the branch.

> Now, using messagelog for buffering the forwarded requests eliminates dandling goroutines when the destination replica stops receiving the messages. However, this will still accumulate the forwarded messages in the unicast message logs, whereas we only need to keep at most one request per client (until we add support for concurrent client requests). I think we can try to improve this as a separate issue.

I agree, I added an inline comment to mention this.

> I suggest passing forwardChan to makeMessageSupplier instead. This way would better align with what we already do with the common message log.

Applied the suggestion, thank you.

core/message-handling.go
```diff
@@ -253,14 +253,20 @@ func makeMessageStreamHandler(handle incomingMessageHandler, logger *logging.Log
...
 	for peerID := uint32(0); peerID < n; peerID++ {
 		if peerID == replicaID {
 			continue
 		}
 
+		// TODO: using messagelog for buffering the forwarded requests could
+		// make the requests accumulated when the destination replica stops
+		// receiving them. Currently that does not happen because we only need
+		// to keep at most one request per client. We will need some update when
+		// we add support for concurrent client requests.
```
sergefdrv: Sorry for confusion, but indefinitely many Request messages can accumulate there, even though we currently don't support concurrent requests from the same client. This is because each _new_ Request may be added to the unicast message log for forwarding but never removed from there (check `core/internal/clientstate/request-seq.go`). 

I was thinking of a new component to replace the unicast message logs here. The new component would act as an asynchronous buffer, limited to one Request message per client. Maybe there are other possible solutions. Maybe we should update this comment, merge the PR and tackle the remaining problem as a separate issue.

sergefdrv (changes requested):

core/message-handling.go
nhoriguchi: Thanks for the explanation, I updated the comment (pushed to the branch):

https://github.com/Naoya-Horiguchi/minbft/blob/683019e6dd07b1bb6d05ccb55b932fb17bbb24cc/core/message-handling.go#L262-L265

> I was thinking of a new component to replace the unicast message logs here. The new component would act as an asynchronous buffer, limited to one Request message per client. Maybe there are other possible solutions. 

I don't have any specific idea for now, so I'll look into "new component" approach first.

sergefdrv (approved):

PR #169, opened 2020-04-12
Introduce tests for a few Byzantine faulty cases

nhoriguchi:
This pull request is trying to add test code for a few simple Byzantine faulty cases. As discussed in #127, we could have a hybrid approach of the complementary two approaches:

- (1) tests whole network's behavior by mimicking all replicas individually, and
- (2) tests one tested replica by mimicking surrounding network including the all other replicas.

This PR make a step on approach (1), and some of the simplest cases will be covered. Please see individual patches for more details.

One thing I'd like to discuss is that I don't think that inserting some hook into main logic like below is clean and optimal.

https://github.com/Naoya-Horiguchi/minbft/blob/94dfc5059779f71c2df58efaa411d597139beb94/core/message-handling.go#L256-L268

One possible better way to me is to convert package method `startPeerConnections` into instance method  of `replica struct`, then it's overwritable by introducing another variant struct like `replicaBftTest struct` which embeds the original `replica struct`. This approach seems better in the way that we can clearly separate the test logic from the main logic, and in another way that we could generalize the approach to other functions components in minbft package. One downside is that the code need to be changed in design level and the cost of the change might not be small.

I'm glad if I can hear about any ideas.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -50,6 +50,13 @@ type Configer interface {
 
 	// starts when sends VIEW-CHANGE and stops when receives a valid NEW-VIEW
 	TimeoutViewChange() time.Duration
+
+	// returns true if replica @replicaID is configured to ignore
+	// (i.e. refuse to send any message to) replica @peerID via
+	// replica setting debug.selectiveignorereplica.
+	//
+	// This interface is supposed to used only for debug purpose.
+	SelectiveIgnorantReplicas(replicaID, peerID uint32) bool
 }
 
 //======= Interface for module 'network' =======
--- a/api/mocks/mock.go
+++ b/api/mocks/mock.go
@@ -90,6 +90,20 @@ func (mr *MockConfigerMockRecorder) N() *gomock.Call {
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "N", reflect.TypeOf((*MockConfiger)(nil).N))
 }
 
+// SelectiveIgnorantReplicas mocks base method
+func (m *MockConfiger) SelectiveIgnorantReplicas(arg0, arg1 uint32) bool {
+	m.ctrl.T.Helper()
+	ret := m.ctrl.Call(m, "SelectiveIgnorantReplicas", arg0, arg1)
+	ret0, _ := ret[0].(bool)
+	return ret0
+}
+
+// SelectiveIgnorantReplicas indicates an expected call of SelectiveIgnorantReplicas
+func (mr *MockConfigerMockRecorder) SelectiveIgnorantReplicas(arg0, arg1 interface{}) *gomock.Call {
+	mr.mock.ctrl.T.Helper()
+	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "SelectiveIgnorantReplicas", reflect.TypeOf((*MockConfiger)(nil).SelectiveIgnorantReplicas), arg0, arg1)
+}
+
 // TimeoutPrepare mocks base method
 func (m *MockConfiger) TimeoutPrepare() time.Duration {
 	m.ctrl.T.Helper()
--- a/core/bft_test.go
+++ b/core/bft_test.go
@@ -1,0 +1,98 @@
+// Copyright (c) 2020 NEC Solution Innovators, Ltd.
+//
+// Authors: Naoya Horiguchi <naoya.horiguchi@nec.com>
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package minbft_test
+
+import (
+	"bytes"
+	"fmt"
+	"testing"
+	"text/template"
+
+	"github.com/hyperledger-labs/minbft/sample/config"
+)
+
+const cfgBftTemplate = `
+protocol:
+  "n": {{.N}}
+  f: {{getF .N}}
+  checkpointPeriod: 10
+  logsize: 20
+  timeout:
+    request: 2s
+    viewchange: 3s
+debug:
+  selectiveignorantreplicas:
+    {{range $k, $v := .Ignore}}{{$k}}:{{range $s := $v}} {{$s}}{{end}}
+    {{end}}
+`
+
+// createTestCfgBft creates config file for Byzantine faulty cases.
+func createTestCfgBft(numReplica int, ignore map[string][]int) []byte {
+	var err error
+	var testCfgBft bytes.Buffer
+
+	t := template.New("cfgBftTemplate")
+	t = t.Funcs(template.FuncMap{
+		"getF": func(n int) int { return (n - 1) / 2 },
+	})
+	t = template.Must(t.Parse(cfgBftTemplate))
+	if err = t.Execute(&testCfgBft, struct {
+		N      int
+		Ignore map[string][]int
+	}{
+		N:      numReplica,
+		Ignore: ignore,
+	}); err != nil {
+		panic(err)
+	}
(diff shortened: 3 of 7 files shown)


ynamiki:
Just for your information, I'd like to share how the PBFT implementation in Fabric v0.6.1 is tested. It has a parameter in a configuration file as we do, to cause byzantine behavior; a peer will not send a broadcast message to one of peers with 1/3 possibility. This code is written in a production file (not a test file).

https://github.com/hyperledger/fabric/blob/e728c5c22160620e189e85be6becb7cbf75d87dc/consensus/pbft/pbft-core.go#L1286

> One possible better way to me is to convert package method startPeerConnections into instance method of replica struct, then it's overwritable by introducing another variant struct like replicaBftTest struct which embeds the original replica struct. This approach seems better in the way that we can clearly separate the test logic from the main logic, and in another way that we could generalize the approach to other functions components in minbft package. One downside is that the code need to be changed in design level and the cost of the change might not be small.

The first way in looks good for me but I am not sure it is the best design in Go. As you mentioned, the way needs many changes and I think we can proceed with the current design because there is only one paremeter at now.

ynamiki (commented):
Just for your information, I'd like to share how the PBFT implementation in Fabric v0.6.1 is tested. It has a parameter in a configuration file as we do, to cause byzantine behavior; a peer will not send a broadcast message to one of peers with 1/3 possibility. This code is written in a production file (not a test file).

https://github.com/hyperledger/fabric/blob/e728c5c22160620e189e85be6becb7cbf75d87dc/consensus/pbft/pbft-core.go#L1286

> One possible better way to me is to convert package method startPeerConnections into instance method of replica struct, then it's overwritable by introducing another variant struct like replicaBftTest struct which embeds the original replica struct. This approach seems better in the way that we can clearly separate the test logic from the main logic, and in another way that we could generalize the approach to other functions components in minbft package. One downside is that the code need to be changed in design level and the cost of the change might not be small.

The first way in looks good for me but I am not sure it is the best design in Go. As you mentioned, the way needs many changes and I think we can proceed with the current design because there is only one paremeter at now.

nhoriguchi:
> Just for your information, I'd like to share how the PBFT implementation in Fabric v0.6.1 is tested. It has a parameter in a configuration file as we do, to cause byzantine behavior; a peer will not send a broadcast message to one of peers with 1/3 possibility. This code is written in a production file (not a test file).
> 
> https://github.com/hyperledger/fabric/blob/e728c5c22160620e189e85be6becb7cbf75d87dc/consensus/pbft/pbft-core.go#L1286

Thanks for sharing/checking, Namiki-san. So such a faulty case seems possible way to go. In order to test other Byzantine cases in more systematic manner, I think that we have to understand what kind of Byzantine faults we can have. What I think of now are categorized as below:
- selective ignorance: a Byzantine faulty replica refuses to send only to a set of other replicas. "Intentionally delay message" case may be included into this category.
- inconsistent response: a Byzantine faulty replica responds to a set of replicas as the protocol specifies, but responds inconsisted (or manipulated) messages to the other replicas.
- unexpected message: a Byzantine faulty replica sends random messages to one or more replicas at random.
- (we could have better categorization...)

"selective ignorance" is relatively easy to test, but other types are protocol specific and maybe hard to test in current approach.

sergefdrv:
I am really sorry for such late feedback...

> One thing I'd like to discuss is that I don't think that inserting some hook into main logic [...] is clean and optimal.

I believe even the core logic of correct behavior alone is already substantially complex, so inserting hooks for Byzantine behavior would clutter the code and make it even harder to follow, as well as introduce some additional performance overhead.

> One possible better way to me is to convert package method `startPeerConnections` into instance method of `replica struct`, then it's overwritable by introducing another variant struct like `replicaBftTest struct` which embeds the original `replica struct`. This approach seems better in the way that we can clearly separate the test logic from the main logic, and in another way that we could generalize the approach to other functions components in minbft package. 

Maybe we could convert parts of connectivity-related code (it's not much) into dependency-injection style, similar to the core protocol logic which is constructed from smaller components, mostly using function closure technique. It would help if we make `defaultIncomingMessageHandler` function more modular. That would allow us to reuse existing components for correct behavior and construct Byzantine replica instances by replacing individual components, clearly separating normal and testing logic.

Another approach would be to describe Byzantine replica behavior in specific test scenarios. This would require some testing framework to imitate Byzantine replica by merely executing/interpreting the scenario, which is basically generating and sending messages. The same approach would help us to test individual replicas by imitating the surrounding network.

As a side note, formal verification could be another possibility, but as far as I know, this is not really an option with imperative languages like Go.

> [...] In order to test other Byzantine cases in more systematic manner, I think that we have to understand what kind of Byzantine faults we can have. What I think of now are categorized as below:
> 
> * selective ignorance: a Byzantine faulty replica refuses to send only to a set of other replicas. "Intentionally delay message" case may be included into this category.
> 
> * inconsistent response: a Byzantine faulty replica responds to a set of replicas as the protocol specifies, but responds inconsistent (or manipulated) messages to the other replicas.
> 
> * unexpected message: a Byzantine faulty replica sends random messages to one or more replicas at random.

Forged message or "internally inconsistent" messages (e.g. `Prepare` from backup replica) would fall under "unexpected message" category, right?

> "selective ignorance" is relatively easy to test, but other types are protocol specific and maybe hard to test in current approach.

In fact, "selective" ignorance can be simulated in a specifically crafted implementation of `ReplicaConnector`, completely separated from the core, but other types of failure would require somewhat more powerful mechanisms...

sergefdrv (commented):
Please check my feedback in https://github.com/hyperledger-labs/minbft/pull/169#issuecomment-624093841. Besides that, I think we might already like to apply changes from 4b09c012f480bcec6adc1090990a3e06cf629664, in a separate PR.

nhoriguchi:
> I believe even the core logic of correct behavior alone is already substantially complex, so inserting hooks for Byzantine behavior would clutter the code and make it even harder to follow, as well as introduce some additional performance overhead.

Yes, so we would want some way to avoid this.

> Maybe we could convert parts of connectivity-related code (it's not much) into dependency-injection style, similar to the core protocol logic which is constructed from smaller components, mostly using function closure technique. It would help if we make `defaultIncomingMessageHandler` function more modular. That would allow us to reuse existing components for correct behavior and construct Byzantine replica instances by replacing individual components, clearly separating normal and testing logic.

OK, so this approach seem to be promising.

> As a side note, formal verification could be another possibility, but as far as I know, this is not really an option with imperative languages like Go.
> 
> > [...] In order to test other Byzantine cases in more systematic manner, I think that we have to understand what kind of Byzantine faults we can have. What I think of now are categorized as below:
> > 
> > * selective ignorance: a Byzantine faulty replica refuses to send only to a set of other replicas. "Intentionally delay message" case may be included into this category.
> > * inconsistent response: a Byzantine faulty replica responds to a set of replicas as the protocol specifies, but responds inconsistent (or manipulated) messages to the other replicas.
> > * unexpected message: a Byzantine faulty replica sends random messages to one or more replicas at random.
> 
> Forged message or "internally inconsistent" messages (e.g. `Prepare` from backup replica) would fall under "unexpected message" category, right?

In my opinion, Byzantine behavior in "inconsistent response" category is something predictable and maybe it could be simulated by introducing a variant with some small changes from the normal core logic.  On the other hand, I assumed in "unexpected message" category that a testing replica receives messages in somehow unpredictable manner (meaning very far from our protocol).  The latter ones are by nature harder to define but maybe the above-mentioned scenario based approach is suitable for simulation.

> 
> > "selective ignorance" is relatively easy to test, but other types are protocol specific and maybe hard to test in current approach.
> 
> In fact, "selective" ignorance can be simulated in a specifically crafted implementation of `ReplicaConnector`, completely separated from the core, but other types of failure would require somewhat more powerful mechanisms...

Sure, I first thought of adding another `ReplicaConnector` implementation for testing, but I felt difficulty (I don't remember details of it, sorry) and didn't make it.

> Please check my feedback in [#169 (comment)](https://github.com/hyperledger-labs/minbft/pull/169#issuecomment-624093841). Besides that, I think we might already like to apply changes from [4b09c01](https://github.com/hyperledger-labs/minbft/commit/4b09c012f480bcec6adc1090990a3e06cf629664), in a separate PR.

OK, I'll submit it later.

nhoriguchi:
I'm implementing the same problem with another approach in #196. I think that it completely includes the testcases covered by this PR, so I no longer have to push this PR for merging. I close this issue.

PR #175, opened 2020-05-04
Reverse peer message stream

### LINKED ISSUE #174: Blocking message stream on handling forwarded Request
I am checking the message handling code and have noticed the following. Currently, we don't distinguish whether an incoming connections originates from a client or from another replica: both message streams are eventually handled by the function constructed in `makeMessageStreamHandler`. This function will try to deliver any message produced by the supplied `incomingMessageHandler` as a reply to the handled incoming message. In case of `Request` message, it will return a non-nil `replyChan` and send there a `Reply` message once it is generated.

This works fine when a replica receives a `Request` message form a client, since the client will wait for the `Reply` message on the reply channel. Now we have added request forwarding, a replica might receive `Request` messages forwarded by other replicas, as well. In that case, it will also try supplying the `Reply` message back. However, replicas do not expect any message in reply, so it will block on trying to send the `Reply` message.

@Naoya-Horiguchi When you tested request forwarding, have you experienced any anomaly that would confirm the described problem? I am currently working on changing the message exchange scheme for some other reason. Once I finish, I hope, this issue will be solved, as well. As a quick kludge, we could modify `startPeerConnection` to receive and discard messages from the channel returned by `peerConnector`.

sergefdrv:
This pull request is aimed to improve message exchange between replicas. Please check individual commit messages for more details.

Resolves: #174


nhoriguchi (commented):
Thank you for the work. I tested and confirmed that this change set solves the malfunction mentioned in #174. If I read correctly, this fix comes with commit [c6cb3c](https://github.com/hyperledger-labs/minbft/pull/175/commits/c6cb3ccf8cd6a9f77adf979454755a1c17f0557e) which makes peer check returned `replyChan` and `new` independently. And commit [73dc78](https://github.com/hyperledger-labs/minbft/pull/175/commits/73dc78441973a443350652eed58ef1d4601793db) chances to choose message handlers by stream type. Forwarded REQUESTs are passed through peer message handler which never returns non-nil `replyChan`, so this convinces me that issue #174 is solved.

A few questions:

- commit [acbb00](https://github.com/hyperledger-labs/minbft/pull/175/commits/acbb002a9b2004cd74276d7b5d8fc850237a3e38) states "No correct replica/client should ever send an incorrect message", but here don't we have to consider that messages could be corrupted for example by one-off errors (some transient network errors or memory errors in my mind)? I thought that we should somehow make peer tolerant with such errors, but we don't support reconnection yet, so it might be OK to consider when implementing it (Starting with simple implementation is fine to me.)

- commit [59dfcc](https://github.com/hyperledger-labs/minbft/pull/175/commits/59dfcce51a15d3d138f532d342d3ec96e49d6e33) explains that shifting from "push" model to "pull" model helps to simplify identifying peer message source, but what does "identifying peer message source" mean? I first though of verifying signatures, but HELLO message is not signed, so maybe this seems to have other meanings.

sergefdrv:
> [...] commit [c6cb3c](https://github.com/hyperledger-labs/minbft/pull/175/commits/c6cb3ccf8cd6a9f77adf979454755a1c17f0557e) which makes peer check returned `replyChan` and `new` independently

In fact, that commit makes replicas always check both `new` and `replyChan` return values.

> commit [acbb00](https://github.com/hyperledger-labs/minbft/pull/175/commits/acbb002a9b2004cd74276d7b5d8fc850237a3e38) states "No correct replica/client should ever send an incorrect message", but here don't we have to consider that messages could be corrupted for example by one-off errors (some transient network errors or memory errors in my mind)? I thought that we should somehow make peer tolerant with such errors, but we don't support reconnection yet, so it might be OK to consider when implementing it (Starting with simple implementation is fine to me.)

I believe message corruption during network transport can be practically excluded relying on TLS and TCP protocols. This is assumed with "The connection guarantees authenticated destination, eventual delivery, preserving integrity and the original order of messages." in [the API](https://github.com/hyperledger-labs/minbft/blob/9f0b3f20bb377c35989dd51028370490edfb9b03/api/api.go#L69).

Regarding such failures as memory errors or sporadic software bugs, those should fall into the category of Byzantine (arbitrary) failure which are tolerated within a known threshold of `f` by the protocol. Undoubtedly, we need to support a mechanism for replica recovery and rejoining to avoid accumulating the number of failed replicas in the system. This would be a separate feature to work on in future.

>  * commit [59dfcc](https://github.com/hyperledger-labs/minbft/pull/175/commits/59dfcce51a15d3d138f532d342d3ec96e49d6e33) explains that shifting from "push" model to "pull" model helps to simplify identifying peer message source, but what does "identifying peer message source" mean? I first though of verifying signatures, but HELLO message is not signed, so maybe this seems to have other meanings.

"Message source" here should be understood as the remote peer that transmitted the message, as opposed to the replica that originally generated and singed the message. 

After this commit applied, replicas receive core protocol messages through the communication connections _they initiated themselves_. Note that the connection target/destination is authenticated, as documented in [the API](https://github.com/hyperledger-labs/minbft/blob/9f0b3f20bb377c35989dd51028370490edfb9b03/api/api.go#L69). This makes it possible to attribute any received faulty message to the sender, which is the authenticated connection target, and terminate the connection.

The only message type accepted from unauthenticated _incoming_ peer connection is HELLO message. This message just indicates which messages the sender is interested in, currently based on its replica ID. Thus, malicious replicas could only try to DoS other replicas with HELLO messages which seems quite ineffective, comparing to using a series of normal protocol messages as before. The application/integration could also apply some mechanism to limit the number of allowed connections originating from the same remote peer, further limiting the possibility of DoS attacks.

There would be little benefit in signing HELLO messages because any replica could simply replay HELLO messages received from any other replica. However, note that `PeerMessageStreamHandler` method of [`ConnectionHandler`](https://github.com/hyperledger-labs/minbft/blob/9f0b3f20bb377c35989dd51028370490edfb9b03/api/api.go#L57) (implemented by replica instance) should only be used by the application/integration to serve connections originating form other replicas. It is left to the application which mechanism to use for ensuring that, e.g. mutual TLS, VPN, etc.

nhoriguchi (approved):
> I believe message corruption during network transport can be practically excluded relying on TLS and TCP protocols. This is assumed with "The connection guarantees authenticated destination, eventual delivery, preserving integrity and the original order of messages." in [the API](https://github.com/hyperledger-labs/minbft/blob/9f0b3f20bb377c35989dd51028370490edfb9b03/api/api.go#L69).
> 
> Regarding such failures as memory errors or sporadic software bugs, those should fall into the category of Byzantine (arbitrary) failure which are tolerated within a known threshold of `f` by the protocol. Undoubtedly, we need to support a mechanism for replica recovery and rejoining to avoid accumulating the number of failed replicas in the system. This would be a separate feature to work on in future.

Fair enough, thank you for the clarification.

> 
> > * commit [59dfcc](https://github.com/hyperledger-labs/minbft/pull/175/commits/59dfcce51a15d3d138f532d342d3ec96e49d6e33) explains that shifting from "push" model to "pull" model helps to simplify identifying peer message source, but what does "identifying peer message source" mean? I first though of verifying signatures, but HELLO message is not signed, so maybe this seems to have other meanings.
> 
> "Message source" here should be understood as the remote peer that transmitted the message, as opposed to the replica that originally generated and singed the message.

This definition is what I missed. So the replica which forwards REQUEST is a message source, but not a replica which generated/signed the message.

> ...

All the remaining parts sound convincing to me, so I'm fine to merge this PR.
Many thanks for solving #174.

ynamiki (approved):
I agree with the change of the communication model from the previous conversation. Thanks.

Issue #182, opened 2020-05-07
Update the CI environment


ynamiki:
Before the upcoming release, update the dependencies, the CI environment, and README to use the latest releases:

- [x] Dependencies
- [x] Ubuntu (CI)
- [x] Intel SGX SDK (CI)
- [x] Go (CI)
- [x] GolangCI-Lint (CI)
- [x] README

sergefdrv:
We might also like to check other direct dependencies in `go.mod` for useful updates.

ynamiki:
Done by #185 and #190.

PR #185, opened 2020-05-14
Update the CI environment

ynamiki:
This patch update our CI environment to use the latest versions of:
- Intel SGX SDK (2.9.1)
- GolangCI-Lint (1.27.0)
- Go (1.11 and 1.14)

Note that since Go 1.13, [module-mode is enabled](https://github.com/golang/go/wiki/Modules#go-113) even the project is located inside GOPATH with default `GO111MODULE` setting. So we don't need to set `GO111MODULE=on` explicitly.

This PR is a part of #182.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -1,9 +1,9 @@
 name: Continuous integration
 on: [push, pull_request]
 env:
-  SGX_SDK_URL: https://download.01.org/intel-sgx/linux-2.3.1/ubuntu18.04/sgx_linux_x64_sdk_2.3.101.46683.bin
+  SGX_SDK_URL: https://download.01.org/intel-sgx/sgx-linux/2.9.1/distro/ubuntu18.04-server/sgx_linux_x64_sdk_2.9.101.2.bin
   GOLANGCI_LINT_URL: https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh
-  GOLANGCI_LINT_VERSION: v1.21.0
+  GOLANGCI_LINT_VERSION: v1.27.0
   SGX_MODE: SIM
 
 jobs:
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        go: [1.11, 1.13]
+        go: [1.11, 1.14]
     steps:
       - name: Install Intel SGX SDK
         run: |
@@ -20,7 +20,7 @@ jobs:
           chmod +x sgx_linux_x64_sdk.bin
           echo -e "no\n/opt/intel" | sudo ./sgx_linux_x64_sdk.bin
           rm sgx_linux_x64_sdk.bin
-      - uses: actions/setup-go@v1
+      - uses: actions/setup-go@v2
         with:
           go-version: ${{ matrix.go }}
       - name: Print Go version
@@ -29,7 +29,7 @@ jobs:
       - name: Install GolangCI-Lint
         run: |
           curl -sfL $GOLANGCI_LINT_URL | sudo sh -s -- -b /usr/local/bin $GOLANGCI_LINT_VERSION
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
       - name: Test
         run: |
           source /opt/intel/sgxsdk/environment
--- a/README.md
+++ b/README.md
@@ -117,13 +117,13 @@ sudo apt-get install build-essential pkg-config
 ### Golang ###
 
 Go 1.11 or later is required to build this project (tested with Go 1.11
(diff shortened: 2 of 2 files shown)


ynamiki:
I forgot to sign-off; force-pushed to add it.

nhoriguchi:
I mostly agree with this suggestion.

> This patch update our CI environment to use the latest versions of:
> 
>     * Intel SGX SDK (2.9.1)

Did you test HW mode with latest SGX toolset? This might not be related to CI itself, but early users who read the updated README try to use 2.9.1, so it would be fine to have some brief check.

ynamiki:
@Naoya-Horiguchi Right, I'll check with HW mode.

ynamiki:
The check with HW mode will be done in June.

I update the driver of my SGX-equipped machine, and find that I need to disable Secure Boot through BIOS. Unfortunately I cannot access the BIOS while this COVID-19 lock down.

nhoriguchi:
> The check with HW mode will be done in June.
> 
> I update the driver of my SGX-equipped machine, and find that I need to disable Secure Boot through BIOS. Unfortunately I cannot access the BIOS while this COVID-19 lock down.

OK, so I'll do some check later with SGX enabled Azure VM.

nhoriguchi:
> OK, so I'll do some check later with SGX enabled Azure VM.

As shared in [this gist](https://gist.github.com/Naoya-Horiguchi/9d700ccf328f23791628ce0d374b64b4), the test process is killed in test/usig_test. The testing script is also attached to the gist. I'm not sure why killed, but it might be related that the kernel used in the Azure VM is different from Ubuntu 18.04 original kernel?
~~~
ubuntu@accvm:~/minbft$ uname -r
5.3.0-1020-azure
~~~

ynamiki:
@Naoya-Horiguchi Thanks, I noticed that in [Quickstart](https://docs.microsoft.com/en-us/azure/confidential-computing/quick-create-marketplace#2-install-the-intel-sgx-dcap-driver) by Microsoft use SGX "DCAP driver" while you use "OOT Driver" (sgx_linux_x64_driver_2.6.0_95eaa6f.bin, I found a memo [here](https://download.01.org/intel-sgx/sgx-linux/2.9.1/distro/ubuntu18.04-server/driver_readme.txt)). However I am not sure there difference between OOT (out-of-tree?) and DCAP and whether this is the cause of the issue.

https://download.01.org/intel-sgx/sgx-linux/2.9.1/distro/ubuntu18.04-server/

OOT driver:  sgx_linux_x64_driver_2.6.0_95eaa6f.bin
DCAP driver: sgx_linux_x64_driver_1.33.bin

nhoriguchi:
> @Naoya-Horiguchi Thanks, I noticed that in [Quickstart](https://docs.microsoft.com/en-us/azure/confidential-computing/quick-create-marketplace#2-install-the-intel-sgx-dcap-driver) by Microsoft use SGX "DCAP driver" while you use "OOT Driver" (sgx_linux_x64_driver_2.6.0_95eaa6f.bin, I found a memo [here](https://download.01.org/intel-sgx/sgx-linux/2.9.1/distro/ubuntu18.04-server/driver_readme.txt)). However I am not sure there difference between OOT (out-of-tree?) and DCAP and whether this is the cause of the issue.
> 
> https://download.01.org/intel-sgx/sgx-linux/2.9.1/distro/ubuntu18.04-server/
> 
> OOT driver: sgx_linux_x64_driver_2.6.0_95eaa6f.bin
> DCAP driver: sgx_linux_x64_driver_1.33.bin

You are right, test passed with DCAP driver. Thank you very much.

nhoriguchi (approved):

sergefdrv (approved):
:+1:

nhoriguchi:
BTW, I think that some simple test script for HW mode like [test.sh](https://gist.github.com/Naoya-Horiguchi/9d700ccf328f23791628ce0d374b64b4#file-test-sh) might be worth sharing in the source tree for manual testing. If you like this idea, I'll create a separate PR.

sergefdrv:
> BTW, I think that some simple test script for HW mode like [test.sh](https://gist.github.com/Naoya-Horiguchi/9d700ccf328f23791628ce0d374b64b4#file-test-sh) might be worth sharing in the source tree for manual testing. If you like this idea, I'll create a separate PR.

I think it's a good idea. It looks like the script is mostly about setting up the prerequisites to run `make test` in HW mode, so maybe adjust and name it correspondingly?

ynamiki:
@Naoya-Horiguchi Thanks for the testing!

> BTW, I think that some simple test script for HW mode like test.sh might be worth sharing in the source tree for manual testing. If you like this idea, I'll create a separate PR.

Good idea👍

PR #187, opened 2020-05-21
Update CODEOWNERS

### LINKED ISSUE #183: Code review responsibilities
As we discussed in #12, pull requests must be reviewed by one of the [minbft-committers](https://github.com/orgs/hyperledger-labs/teams/minbft-committers) members (former [nec-blockchain](https://github.com/nec-blockchain) members). So far I used to request review from both @ynamiki and @Naoya-Horiguchi for my pull requests and wait for both to approve before merging. I was wondering if this might be too much overhead and you would like to somehow split review responsibilities?

ynamiki:
Based on the discussion #183, update CODEOWNERS.

@sergefdrv @Naoya-Horiguchi Let's discuss a tangible assignment.

In a CODEOWNERS file, the last matching pattern takes the most precedence. So my initial commit means I take responsibility for everything else (i.e. non-code files) other than listed in the following lines.

https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners

Resolves #183.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,8 @@
-* @sergefdrv @Naoya-Horiguchi @ynamiki
+* @Naoya-Horiguchi @ynamiki
+
+/api/ @sergefdrv @Naoya-Horiguchi
+/client/ @sergefdrv @Naoya-Horiguchi
+/core/ @sergefdrv @Naoya-Horiguchi
+/messages/ @sergefdrv @Naoya-Horiguchi
+/sample/ @sergefdrv @Naoya-Horiguchi
+/usig/ @sergefdrv @Naoya-Horiguchi


nhoriguchi (commented):
Maybe `/tools` belongs to me?

BTW, interestingly code owner works like this :)

image

sergefdrv (commented):
Maybe we should better keep at least two owners for each area? Consider the case when one of the code owners is also the author of changes (which currently happens most of the time). Then the other code owner would be automatically requested for review.

ynamiki:
> Maybe we should better keep at least two owners for each area?

Right, I have updated; assign @sergefdrv and @Naoya-Horiguchi to the core part, @Naoya-Horiguchi and me to others.

sergefdrv (approved):
Looks good, thanks! :+1:

PR #188, opened 2020-05-21
Remove duplicate UI parsing

sergefdrv:
This pull request improves the code base by removing duplicate parsing of message UI. This will also simplify further changes that I am preparing to submit.

Note that, although most of duplicate UI un-/marshalling is eliminated here, it still remains when generating/verifying UI. This is because UI generation/verification uses `api.Authenticator` via the same interface as generating/verifying normal signatures. We might like to improve this further by adjusting the API to operate on `usig.UI` structure directly.

### DIFF
--- a/core/testutils_test.go
+++ b/core/testutils_test.go
@@ -63,3 +63,8 @@ func randOtherReplicaID(id, n uint32) uint32 {
 	offset := rand.Intn(int(n)-1) + 1
 	return (id + uint32(offset)) % n
 }
+
+// randBytes returns a slice of random bytes
+func randBytes() []byte {
+	return []byte{byte(rand.Int())} // nolint:gosec
+}
--- a/core/commit.go
+++ b/core/commit.go
@@ -74,7 +74,7 @@ func makeCommitValidator(verifyUI uiVerifier, validatePrepare prepareValidator) 
 			return fmt.Errorf("Invalid Prepare: %s", err)
 		}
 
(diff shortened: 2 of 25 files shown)


core/testutils_test.go
```diff
@@ -63,3 +63,8 @@ func randOtherReplicaID(id, n uint32) uint32 {
 	offset := rand.Intn(int(n)-1) + 1
 	return (id + uint32(offset)) % n
 }
+
+// randBytes returns a slice of random bytes
+func randBytes() []byte {
+	return []byte{byte(rand.Int())} // nolint:gosec
```
ynamiki, line 69: Don't we need set seed? Otherwise, `rand.Int()` returns same results for each run. I'm not sure this is expected or not.
https://golang.org/pkg/math/rand/

nhoriguchi (approved):
I'm fine with this PR with resolving Namiki-san's comment. Thank you.

core/testutils_test.go
sergefdrv: True, on one hand, this does not give any randomness between test runs, but on the other hand, makes test results more reproducible. I'm not sure what would be better. Maybe it doesn't matter much in our case because of the way we use the random values in testing.

ynamiki: I don't have the answer to the question, too. I'm fine with this. Thanks.

ynamiki (approved):

PR #189, opened 2020-05-21
Detect commitment after view termination

sergefdrv:
This pull request suggests a way to detect a protocol violation when a Byzantine replica continues generating message for a view after it has terminated the view by generating a ViewChange message, as well as to simplify implementation of view change being developed. The proposed change adds an additional requirement on Prepare/Commit message sequence of the same view. It will consequently require to sign Checkpoint messages with a regular signature as opposed to certifying them with USIG. Note that a similar approach was discussed in a paper published in "IEEE Transactions on Computers" journal. Please check commit messages for more details.

### DIFF
--- a/core/commit_test.go
+++ b/core/commit_test.go
@@ -28,6 +28,7 @@ import (
 	"github.com/stretchr/testify/require"
 
 	testifymock "github.com/stretchr/testify/mock"
+	yaml "gopkg.in/yaml.v2"
 
 	"github.com/hyperledger-labs/minbft/messages"
 	"github.com/hyperledger-labs/minbft/usig"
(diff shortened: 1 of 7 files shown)


sergefdrv:
NB: This PR is based on commits from #188.

sergefdrv:
Summary of [changes](https://github.com/hyperledger-labs/minbft/compare/2fe03a632c08a7b3c395dd2ccd59c1ecb8aa6f86..845bc82cfcb081b8cc272a42a8d79d7bb57bf446): clean rebase.

core/commit_test.go
```diff
@@ -221,160 +236,85 @@ func TestMakeCommitmentCollectorConcurrent(t *testing.T) {
...
-			ok:        true,
-			done:      true,
-		}, {
-			desc:      "Extra commitment for the first request",
-			prepareCV: 1,
-			replicaID: 2,
-			ok:        true,
-			done:      true,
-		}},
+func TestMakeCommitmentAcceptor(t *testing.T) {
```
ynamiki, line 239: It would be good if we have error cases (which test that inputs cause an expected error) but it may be future work.

ynamiki (approved):
Looks good for me, thanks.

nhoriguchi (approved):

core/commit_test.go
sergefdrv: I am not sure if I understood exactly, but the `ok` test case parameter requires that either nil or non-nil error must be returned. Thus the test would fail in case of unexpected error.

ynamiki: Sorry, I was wrong. Error cases are tested.

PR #190, opened 2020-05-28
Update dependencies

ynamiki:
This patch updates dependencies to the latest minor versions (i.e. this will not change major versions, which could break compatibility according to the [semver](https://semver.org) scheme).

This is part of #182.


nhoriguchi (approved):
I guess updating protobuf to the newer major version could have some impact on stability because this project heavily depends on the package. So I'm OK to keep using v1 at least until all core features are implemented and stable enough (or we have any good reason.)

sergefdrv (commented):
> This patch updates dependencies to the latest minor versions (i.e. this will not change major versions, which could break compatibility according to the semver scheme).

Why don't we try and switch to the latest major releases of all direct dependencies? 

In particular, we could switch from using the obsolete `github.com/golang/protobuf/proto` to `google.golang.org/protobuf/proto`. I quickly tried and both unit testing and the example seem to work fine.

sergefdrv:
> In particular, we could switch from using the obsolete `github.com/golang/protobuf/proto` to `google.golang.org/protobuf/proto`. I quickly tried and both unit testing and the example seem to work fine.

First, I regenerated the protobuf code using the following tool versions:

     protoc-gen-go v1.23.0
     protoc        v3.12.1

Then, changed import path of `proto` package in `messages/protobuf/pb/utils.go` and `messages/protobuf/impl.go`:

```diff
-	"github.com/golang/protobuf/proto"
+	"google.golang.org/protobuf/proto"
```

sergefdrv:
> I guess updating protobuf to the newer major version could have some impact on stability because this project heavily depends on the package. So I'm OK to keep using v1 at least until all core features are implemented and stable enough (or we have any good reason.)

I guess it is unlikely that the new proto package would be less stable. We are trying to upgrade to `v1.4.2` which should in fact be implemented in terms of the new version, according to [this post](https://blog.golang.org/protobuf-apiv2):

> github.com/golang/protobuf@v1.4.0 is a version of APIv1 implemented in terms of APIv2. The API is the same, but the underlying implementation is backed by the new one.

ynamiki:
Thanks for comments, I will try major updates.

nhoriguchi:
[Off Topic] I feel that I have to withdraw my approval, but there seems no way for it. Enabling "Dismiss stale pull request approvals when new commits are pushed" setting in branch protection tab (see below) would help us, although I don't have permission to change it on this project. If you think it's helpful, let's enable the setting.

image

ynamiki:
> First, I regenerated the protobuf code using the following tool versions:
> 
> ```
>  protoc-gen-go v1.23.0
>  protoc        v3.12.1
> ```
> 
> Then, changed import path of `proto` package in `messages/protobuf/pb/utils.go` and `messages/protobuf/impl.go`:
> 
> ```diff
> -	"github.com/golang/protobuf/proto"
> +	"google.golang.org/protobuf/proto"
> ```

@sergefdrv `make generate` was failed with the following error in my environment with the same versions of `protoc` and `protoc-gen-go`.

```
-go_out: protoc-gen-go: plugins are not supported; use 'protoc --go-grpc_out=...' to generate gRPC
messages/protobuf/pb/generate.go:17: running "protoc": exit status 1
```

Because we have `//go:generate protoc messages.proto --go_out=plugins=grpc:.` to generate gRPC service definitions (`plugins=grpc`) and `protoc-gen-go` v1.20 [does not support](https://github.com/protocolbuffers/protobuf-go/releases#v1.20-grpc-support) it, this is expected results. The document also says that it will be supported in a future release and we need to use `github.com/golang/protobuf` (i.e. old) version of the tool now.

Could you show the way to compile the `.proto` files?

sergefdrv:
> Because we have `//go:generate protoc messages.proto --go_out=plugins=grpc:.` to generate gRPC service definitions (`plugins=grpc`) and `protoc-gen-go` v1.20 [does not support](https://github.com/protocolbuffers/protobuf-go/releases#v1.20-grpc-support) it, this is expected results. The document also says that it will be supported in a future release and we need to use `github.com/golang/protobuf` (i.e. old) version of the tool now.
> 
> Could you show the way to compile the `.proto` files?

Right, `google.golang.org/protobuf/cmd/protoc-gen-go` doesn't support gRPC yet. I used the latest release from https://github.com/golang/protobuf/releases/tag/v1.4.2.

```sh
mkdir -p $GOPATH/src/github.com/golang
cd $_
git clone https://github.com/golang/protobuf.git
cd protobuf
git checkout v1.4.2
cd protoc-gen-go
go install
```

Note that the old package's tool is only used to generate the code, whereas we can use the new version in our code.

sergefdrv:
> I feel that I have to withdraw my approval, but there seems no way for it

Perhaps, re-request review button image in the reviewers section would do the trick.

ynamiki:
@sergefdrv Thanks, the generation is succeeded.

I was confused by the following comment in the generated files (`.pb.go`):
```
// Code generated by protoc-gen-go. DO NOT EDIT.
// versions:
// 	protoc-gen-go v1.24.0
// 	protoc        v3.12.1
```
I need to use `protoc-gen-go` v1.4, not v1.24.0.

sergefdrv:
> Enabling "Dismiss stale pull request approvals when new commits are pushed" setting in branch protection tab (see below) would help us, although I don't have permission to change it on this project. If you think it's helpful, let's enable the setting.

I think this setting could be helpful to avoid confusion.

ynamiki:
@sergefdrv @Naoya-Horiguchi I have updated the patch; now includes not only minor update but also major updates. We have only Protocol Buffers for major updates.

nhoriguchi (approved):
Looks good to me.

sergefdrv (approved):
:+1:

PR #192, opened 2020-05-31
Fix Makefile

ynamiki:
The uninstall target in the Makefile forgot to delete the installed file `libusig_shim.so`.
This patch fix the issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,6 +64,7 @@ install: build
 uninstall:
 	rm -f $(bindir)/keytool
 	rm -f $(bindir)/peer
+	rm -f $(libdir)/libusig_shim.so
 	rm -f $(libdir)/libusig.signed.so
 
 clean: usig-clean


Makefile
```diff
@@ -64,6 +64,7 @@ install: build
 uninstall:
 	rm -f $(bindir)/keytool
 	rm -f $(bindir)/peer
+	rm -f $(libdir)/libusig_shim.so
```
nhoriguchi, line 67: Good. And it seems to me that we could remove `$(libdir)` on target `clean`, could you update it too?

ynamiki: If I understand you correctly, I should update the Makefile as follows:

```
clean: usig-clean
	rm -rf $(builddir)
	-rmdir $(libdir) # Add this line
```

Let me confirm some points:
- Why in the target `clean`? I guess `clean` doesn't modify `install`-ed items generally.
- Why only `$(libdir)`? `install` creates (if needed) two directories: `$(bindir)` and `$(libdir)`. I'm not clear on removing only `$(libdir)`.

I think we could remove `$(bindir)` and `$(libdir)` in the target `uninstall` if the directories are empty. But I did not do because I found that it isn't done in many cases:

>  automake's generated uninstall target does not automatically delete empty directories because it does not know if the the application owns the directories

https://stackoverflow.com/q/19521335

nhoriguchi: > Let me confirm some points:
> 
> * Why in the target `clean`? I guess `clean` doesn't modify `install`-ed items generally.
> * Why only `$(libdir)`? `install` creates (if needed) two directories: `$(bindir)` and `$(libdir)`. I'm not clear on removing only `$(libdir)`.

Ah, my bad... 

> 
> I think we could remove `$(bindir)` and `$(libdir)` in the target `uninstall` if the directories are empty. But I did not do because I found that it isn't done in many cases:
> > automake's generated uninstall target does not automatically delete empty directories because it does not know if the the application owns the directories
> 
> https://stackoverflow.com/q/19521335

OK, so the current patch is good. Thank you for the explanation.

nhoriguchi (approved):

sergefdrv (approved):
:+1:

Issue #163, opened 2020-01-26
Create Docker container images of example applications


ynamiki:
A future release of [Intel SGX](https://01.org/intel-softwareguard-extensions) may introduce a new feature to support for running SGX applications inside Docker containers. The feature described in the What's New section in [Intel SGX 2.8 Release Note](https://01.org/intel-softwareguard-extensions/downloads/intel-sgx-linux-2.8-release) but it seems not to be included in this release and is still an ongoing issue: intel/linux-sgx#486.

Once the feature is released, we will consider providing container images of our example applications so that users can try our code with minimum setup.

ynamiki:
I noticed Intel SGX Linux 2.9.1, which was released in early April brings some documents about Docker even there are no mentions in the release note:
https://github.com/intel/linux-sgx/tree/sgx_2.9.1/docker/build

I'll take a look the document to consider whether we can use it.

Note that, the release updates system requirements; now we need Ubuntu 18.04, and Ubuntu 16.04 is dropped.

ynamiki:
Done in #193 for the simulation mode. Needs more work for the HW mode.

PR #193, opened 2020-06-04
Create a Dockerfile

ynamiki:
A Dockerfile for the example described in README.md in the project root. By using this, a user can try our project without environment setup (i.e. installing SGX SDK, build, etc.). Note that, in this container, we use the SGX simulation mode so an Intel SGX-equipped processor is not required to run.

This is a part of #163. I'll try to create another Dockerfile for the HW mode later in another PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,8 @@ prefix ?= sample
 bindir := $(prefix)/bin
 libdir := $(prefix)/lib
 
-.PHONY: all build install uninstall clean check test lint generate
+.PHONY: all build install uninstall clean check test lint generate  \
+        docker docker-clean
 
 all: build check
 test: check
@@ -41,6 +42,8 @@ help:
 	@echo '  all (default)    - Build and test all'
 	@echo '  build            - Build all'
 	@echo '  install          - Build and install artifacts'
+	@echo '  docker           - Build a Docker image'
+	@echo '  docker-clean     - Remove a Docker image'
 	@echo '  uninstall        - Uninstall artifacts'
 	@echo '  clean            - Remove all build artifacts'
 	@echo '  check|test       - Run all tests'
@@ -60,6 +63,12 @@ install: build
 	$(INSTALL_PROGRAM) -D $(builddir)/peer $(bindir)/peer
 	$(INSTALL_DATA) -D usig/sgx/shim/libusig_shim.so $(libdir)/libusig_shim.so
 	$(INSTALL_DATA) -D usig/sgx/enclave/libusig.signed.so $(libdir)/libusig.signed.so
+
+docker:
+	docker build -f sample/docker/Dockerfile -t minbft .
+
+docker-clean:
+	docker rmi -f minbft
 
 uninstall:
 	rm -f $(bindir)/keytool
--- a/README.md
+++ b/README.md
@@ -8,6 +8,12 @@
   * [What is MinBFT](#what-is-minbft)
   * [Why MinBFT](#why-minbft)
   * [Concepts](#concepts)
+  * [Quick Start](#quick-start)
+      * [Prerequisites](#prerequisites)
+      * [Building a Container Image](#building-a-container-image)
+      * [Running Replicas](#running-replicas)
+      * [Submitting Requests](#submitting-requests)
+      * [Tear Down](#tear-down)
   * [Requirements](#requirements)
       * [Operating System](#operating-system)
       * [Golang](#golang)
@@ -102,6 +108,117 @@ number of consenting nodes than PBFT.
 
 For more detailed description of the protocol, refer to [Efficient
 Byzantine Fault-Tolerance paper][minbft-paper].
+
+## Quick Start ##
+
+This quick start shows you how to run this project using Docker
+containers. This is the easiest way to try out this project with minimal
+setup. If you want to build and run without Docker, skip this section.
+
+### Prerequisites ###
+
+To run the containers, the following software must be installed on your
+system.
+
+ - [Docker Engine](https://docs.docker.com/engine/) (tested with version
+   19.03.6)
+ - [Docker Compose](https://docs.docker.com/compose/) (tested with
+   version 1.17.1)
+
+If you are using Ubuntu 18.04, they can be installed as follows:
+
+```sh
+sudo apt-get install docker.io docker-compose
+```
+
+Note that SGX-enabled CPU is not required to run the containers; they
+will run in the simulation mode provided by Intel SGX SDK. We plan to
+have another container image which runs in the HW mode (i.e. using
+"real" hardware features) in a future release.
+
+### Building a Container Image ###
+
+Build an image of the containers as follows:
+
+```sh
+sudo docker build -f sample/docker/Dockerfile -t minbft .
+```
+
+Note that, by defaut, the `docker` command needs to be executed by
+`root`. Refer to [Docker's
+document](https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user)
+for details.
+
+If your system has the `make` command, the following command also can be
+used.
+
+```sh
+sudo make docker
+```
+
+### Running Replicas ###
+
+To start up an example consensus network of replicas, invoke the
+following commands:
+
+```sh
+sudo UID=$UID docker-compose -f sample/docker/docker-compose.yml up -d
+```
+
+This will start the replica nodes as 3 separate containers in
+background.
+
+### Submitting Requests ###
+
+Requests can be submitted for ordering and execution to the example
+consensus network as follows:
+
+```sh
+sudo docker-compose -f sample/docker/docker-compose.yml \
+  run client request "First request" "Second request" "Another request"
+```
+
+This command should produce the following output showing the result
+of ordering and execution of the submitted requests:
+
+```
+Reply: {"Height":1,"PrevBlockHash":null,"Payload":"Rmlyc3QgcmVxdWVzdA=="}
+Reply: {"Height":2,"PrevBlockHash":"DuAGbE1hVQCvgi+R0E5zWaKSlVYFEo3CjlRj9Eik5h4=","Payload":"U2Vjb25kIHJlcXVlc3Q="}
+Reply: {"Height":3,"PrevBlockHash":"963Kn659GbtX35MZYzguEwSH1UvF2cRYo6lNpIyuCUE=","Payload":"QW5vdGhlciByZXF1ZXN0"}
+```
+
+The output shows the submitted requests being ordered and executed by a
+sample blockchain service. The service executes request by simply
+appending a new block for each request to the trivial blockchain
+maintained by the service.
+
+### Tear Down ###
+
+The following command can be used to terminate running containers:
+
+```sh
+sudo docker-compose -f sample/docker/docker-compose.yml down
+```
+
+The containers create some files while running. These files can be
+deleted as follows.
+
+```sh
+rm -f sample/docker/keys.yaml sample/docker/.keys.yaml.lock
+```
+
+The Docker image can be deleted as follow.
+
+```sh
+sudo docker rmi minbft
+```
+
+If your system has the `make` command, the following command also can be
+used.
+
+```sh
+sudo make docker-clean
+```
 
 ## Requirements ##
 
--- a/sample/docker/Dockerfile
+++ b/sample/docker/Dockerfile
@@ -1,0 +1,35 @@
+FROM ubuntu:18.04 AS build
+COPY --from=golang:1.14.4-buster /usr/local/go /usr/local/go/
+ENV PATH /usr/local/go/bin:$PATH
+ENV SGX_SDK_URL https://download.01.org/intel-sgx/sgx-linux/2.9.1/distro/ubuntu18.04-server/sgx_linux_x64_sdk_2.9.101.2.bin
+RUN apt-get update && apt-get install -y --no-install-recommends \
+		build-essential \
+		ca-certificates \
+		pkg-config \
+		python \
+		wget \
+	&& rm -rf /var/lib/apt/lists/* \
+	&& wget -q -O sgx_linux_x64_sdk.bin $SGX_SDK_URL \
+	&& chmod +x sgx_linux_x64_sdk.bin \
+	&& echo -e "no\n/opt/intel" | ./sgx_linux_x64_sdk.bin
+COPY . minbft/
+ENV SGX_MODE SIM
+RUN cd minbft \
+	&& . /opt/intel/sgxsdk/environment \
+	&& make prefix=/opt/minbft install
+
+FROM ubuntu:18.04
+RUN apt-get update && apt-get install -y --no-install-recommends \
+		libssl1.1 \
+	&& rm -rf /var/lib/apt/lists/*
+COPY --from=build /opt/intel/sgxsdk /opt/intel/sgxsdk/
+WORKDIR /opt/minbft
+COPY --from=build /opt/minbft ./
+ENV PATH /opt/minbft/bin:$PATH
+ENV LD_LIBRARY_PATH /opt/minbft/lib:$LD_LIBRARY_PATH
+COPY sample/config/consensus.yaml sample/peer/peer.yaml ./
+RUN sed -i 's/:800\([0-2]\)/replica\1:8000/' consensus.yaml
+COPY sample/docker/docker-entrypoint.sh /usr/local/bin/
+VOLUME ["/data"]
+EXPOSE 8000
+ENTRYPOINT ["docker-entrypoint.sh"]
--- a/sample/docker/docker-compose.yml
+++ b/sample/docker/docker-compose.yml
@@ -1,0 +1,23 @@
+version: "3"
+services:
+  replica0: &peer
+    container_name: replica0
+    image: minbft
+    volumes:
+      - ".:/data"
+    expose:
+      - "8000"
+    command: run 0
+    user: ${UID:-1000}
+  replica1:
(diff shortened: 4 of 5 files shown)


ynamiki:
I am thinking of modifying this commit to make a container for each replica node (`peer` process).  Having multiple process in a single container is not good practice.

README.md
```diff
@@ -105,6 +105,9 @@ Byzantine Fault-Tolerance paper][minbft-paper].
 
 ## Requirements ##
 
+Note: We have [a Docker image](sample/docker/README.md) to try this
+project quickly.
```
nhoriguchi: This looks incorrect a little because `sample/docker/` provides a Dockerfile, a script file, and some setting files, not a docker image itself.

sample/docker/README.md
```diff
@@ -0,0 +1,29 @@
+# Docker Image #
+
+## Prerequisites ##
+
+To try this, be sure [Docker](https://www.docker.com) (Docker Engine
+and Docker Compose) are installed on your system.
+
+Note that a host do not need to have Intel SGX; the program runs with
+the simulation mode.
```
nhoriguchi: I found a typo "s/do/does/". And this note seems to be a temporary one, so it might be better to add some words like "for now" or "at this point".

sample/docker/README.md
```diff
@@ -0,0 +1,29 @@
+# Docker Image #
```
nhoriguchi: Maybe the title can be more descriptive like "Running sample application as Docker containers".

sample/docker/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Prerequisites ##
+
+To try this, be sure [Docker](https://www.docker.com) (Docker Engine
+and Docker Compose) are installed on your system.
```
nhoriguchi: This is the first time we mention docker in this project, so it would be kind to add some basic notes on docker (I mean for example that "you need to run this as root with `sudo`", or "you have to check that docker service is running" or something like that) or to point to some references (Hyperledger Fabric [does this](https://hyperledger-fabric.readthedocs.io/en/release-2.0/prereqs.html#docker-and-docker-compose)). With that, we can avoid questions about docker.

sample/docker/docker-compose.yml
```diff
@@ -0,0 +1,21 @@
+version: "3.4"
+services:
+  replica0: &peer
```
nhoriguchi, line 3: Could you give simpler container names with `container_name: replicaN` (default name is `docker_replicaN_1`)? That's a minor thing, but would make containers easier to handle.

nhoriguchi (commented):
I confirmed that the sample application properly ran in Docker container, that's nice.

One concern is that you generate copies of YAML files for docker, which might be a burden from maintenance viewpoint (imagine that if we add some new config parameters, then we might have to update two places). I think that we can use [viper](https://github.com/spf13/viper)'s function to override configs from files with environment variables, and my brief check showed that `keys` in `peer.yaml` can be overridden by environment variable `PEER_KEYS`. Unfortunately, `peers` in `consensus.yaml` is loaded separately from `peer.yaml` so we need change application code to override it with environment variable. So I don't think this should be done in this PR.

README.md
sergefdrv: Maybe just mention that the following requirements apply to native build, and suggest using Docker to get started without worrying about the dependencies?

README.md
```diff
@@ -161,6 +164,9 @@ export SGX_MODE=SIM
 
 ## Getting Started ##
 
+Note: We have [a Docker image](sample/docker/README.md) to try this
+project quickly.
+
```
sergefdrv: Maybe just mention that the following steps apply to native build, and suggest trying out Docker to get started easily? Maybe we could even make it the default option at once?

sample/docker/Dockerfile
```diff
@@ -0,0 +1,38 @@
+FROM ubuntu:18.04 AS sgx-sdk
+RUN apt-get update && apt-get install -y --no-install-recommends \
+		build-essential \
+		ca-certificates \
+		python \
+		wget \
+	&& rm -rf /var/lib/apt/lists/* \
+	&& wget -q -O sgx_linux_x64_sdk.bin https://download.01.org/intel-sgx/sgx-linux/2.9.1/distro/ubuntu18.04-server/sgx_linux_x64_sdk_2.9.101.2.bin \
```
sergefdrv: Maybe we should consider putting the exact URL in an `ENV` instruction:

```dockerfile
ENV SGX_SDK_URL https://download.01.org/intel-sgx/sgx-linux/2.9.1/distro/ubuntu18.04-server/sgx_linux_x64_sdk_2.9.101.2.bin
```

sample/docker/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+	&& chmod +x sgx_linux_x64_sdk.bin \
+	&& echo -e "no\n/opt/intel" | ./sgx_linux_x64_sdk.bin \
+	&& rm sgx_linux_x64_sdk.bin \
+	&& apt-get purge -y --auto-remove \
+		ca-certificates \
+		wget
```
sergefdrv: Maybe we don't need to worry about removing any extra package. Since we use multi-stage docker build, we could simply copy the relevant files from the resulting image and let docker recycle the rest.

sample/docker/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+FROM sgx-sdk AS build
```
sergefdrv: Could we use `ubuntu:18.04` as the base image and simply copy `/opt/intel` from `sgx-sdk` stage?

sample/docker/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+RUN apt-get update && apt-get install -y --no-install-recommends \
+		ca-certificates \
+		pkg-config \
+		wget \
+	&& rm -rf /var/lib/apt/lists/* \
+	&& wget -q -O - https://dl.google.com/go/go1.14.3.linux-amd64.tar.gz | tar -C /usr/local -xzf -
```
sergefdrv: Could we copy `usr/local/go` from the official Go image instead?

sample/docker/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+ENV PATH /usr/local/go/bin:$PATH
+COPY . minbft/
+RUN cd minbft \
+	&& . /opt/intel/sgxsdk/environment \
+	&& SGX_MODE=SIM make prefix=/opt/minbft install
```
sergefdrv: Maybe we should set `SGX_MODE` in an `ENV` instruction.

sample/docker/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+FROM sgx-sdk AS peer
```
sergefdrv: Could we use `ubuntu:18.04` (or even `scratch`) as the base image and just copy the required libraries from `sgx-sdk` stage?

sample/docker/README.md
sergefdrv: It could also be a concise title followed by some short description.

sample/docker/README.md
```diff
@@ -0,0 +1,29 @@
+# Docker Image #
+
+## Prerequisites ##
+
+To try this, be sure [Docker](https://www.docker.com) (Docker Engine
```
sergefdrv: Maybe it would be better to refer to https://docs.docker.com/ since it directly shows "Download and Install" link.

sergefdrv: It could also make sense to clarify the minimal required version.

sample/docker/docker-entrypoint.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+
+. /opt/intel/sgxsdk/environment
+
+if [ $# -lt 2 ]; then
```
sergefdrv: This will not work well with `request` command that receives requests from the standard input.

sample/docker/docker-entrypoint.sh
```diff
@@ -0,0 +1,18 @@
...
+# Only the replica node 0 generates keys.yaml
+if [ "$1$2" = "run0" ]; then
+    keytool -o /data/keys.yaml generate -u lib/libusig.signed.so
+else
+    until [ -f /data/keys.yaml ]; do
+        sleep 1
+    done
+fi
+
+peer "$@"
```
sergefdrv: It is suggested use `exec` so that the process becomes the container’s PID 1. This allows it to receive any Unix signals sent to the container.

sample/docker/docker-entrypoint.sh
```diff
@@ -0,0 +1,18 @@
...
+
+if [ $# -lt 2 ]; then
+    exit
+fi
+
+# Only the replica node 0 generates keys.yaml
+if [ "$1$2" = "run0" ]; then
+    keytool -o /data/keys.yaml generate -u lib/libusig.signed.so
+else
+    until [ -f /data/keys.yaml ]; do
```
sergefdrv: Maybe we could use `flock` command to synchronize one-time initialization of `keys.yaml` file. Something like this should work:

```sh
(
  flock 3
  if [ -f /data/keys.yaml ]; then exit; fi
  keytool -o /data/keys.yaml generate -u lib/libusig.signed.so
  rm -f /data/.keys.yaml.lock
) 3>/data/.keys.yaml.lock
```

sergefdrv (commented):
Thanks for great work. I particularly like the way the docker file is structured with multi-stage build. I think we could also consider adding top-level make targets to build docker images and use those in the example as future work.

README.md
nhoriguchi: You're right, I was short-sighted. Please ignore this comment.

sample/docker/docker-entrypoint.sh
ynamiki: @sergefdrv After some searching, I find an article that says we should not delete the lock file:
https://unix.stackexchange.com/a/368167

I think it is reasonable and plan to add description to remove the lock file manually in the document. What do you think?

sergefdrv: Maybe testing for `/data/keys.yaml` in the critical section would still keep it safe, even if the lock file is removed after generating the key file. But I guess it would not harm if we simply leave the lock file there. Perhaps, it is not worth mentioning in the documentation.

ynamiki:
@Naoya-Horiguchi @sergefdrv Thanks for your many suggestions. I have update the code as well as the document. I the latest commit, I deleted `sample/docker/README.md` and move the contents to an new section "Quick Start" of `README.md` at the root with revising based on your comments.

ynamiki:
@Naoya-Horiguchi I understand your concern. In the last commit, I tried avoiding the duplication of the config files by creating `consensus.yaml` dynamically; [rewriting](https://github.com/hyperledger-labs/minbft/pull/193/commits/6823f9522b4bc7a3c70d82f1858fa4e76b941a29#diff-ab73ca35b6b55a756b36f39c38cbe1caR31) by `sed` based on the original `sample/config/consensus.yaml`.

> One concern is that you generate copies of YAML files for docker, which might be a burden from maintenance viewpoint (imagine that if we add some new config parameters, then we might have to update two places). I think that we can use viper's function to override configs from files with environment variables, and my brief check showed that keys in peer.yaml can be overridden by environment variable PEER_KEYS. Unfortunately, peers in consensus.yaml is loaded separately from peer.yaml so we need change application code to override it with environment variable. So I don't think this should be done in this PR.

README.md
```diff
@@ -8,6 +8,7 @@
   * [What is MinBFT](#what-is-minbft)
   * [Why MinBFT](#why-minbft)
   * [Concepts](#concepts)
+  * [Quick Start](#quick-start)
```
nhoriguchi, line 11: Could you add the links for subsections?

README.md
```diff
@@ -103,6 +104,105 @@ number of consenting nodes than PBFT.
 For more detailed description of the protocol, refer to [Efficient
 Byzantine Fault-Tolerance paper][minbft-paper].
 
+## Quick Start ##
+
+This quick start shows you how to runnning this project using Docker
```
nhoriguchi: s/running/run/

README.md
```diff
@@ -103,6 +104,105 @@ number of consenting nodes than PBFT.
...
+containers. This is the easiest way to try out this project with minimal
+setup. If you want to build and run without Docker, skip this section.
```
nhoriguchi, line 116: (EDITED) Actually I like putting this new section subtree just before "Roadmap" section and inserting short notes as you did in the previous version. But this might conflict with Sergey's suggestion, so I'm fine if you decide to go with this.

README.md
```diff
@@ -103,6 +104,105 @@ number of consenting nodes than PBFT.
...
+Note that, by defaut, the `docker` command needs to be executed by
+`root`. Refer to [Docker's
+document](https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user)
+for details.
+
+If your system have the `make` command and the current user can run the
+`docker` command, the following command also can be used:
+
+```sh
+make docker
```
nhoriguchi: Do we need `sudo` here, too?

README.md
```diff
@@ -103,6 +104,105 @@ number of consenting nodes than PBFT.
...
+```sh
+sudo docker build -f sample/docker/Dockerfile -t minbft .
+```
+
+Note that, by defaut, the `docker` command needs to be executed by
+`root`. Refer to [Docker's
+document](https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user)
+for details.
+
+If your system have the `make` command and the current user can run the
```
sergefdrv: s/have/has/

README.md
sergefdrv: It should be fine because the previous paragraph reads: "if ..., and the current user can run `docker` command".

README.md
```diff
@@ -103,6 +104,105 @@ number of consenting nodes than PBFT.
...
+
+The following command can be used to terminate running containers:
+
+```sh
+sudo docker-compose -f sample/docker/docker-compose.yml down
+```
+
+The containers create some files while running. These files can be
+deleted as follows. Note that the files are created by `root` through
+the container, so `sudo` is required.
```
sergefdrv: We can make the container create files owned by the current user with `docker run ... -u $(id -u):$(id -g) ...` which would roughly map to the following in a Docker Compose file:

```yaml
user: ${UID}
```

Makefile
```diff
@@ -41,6 +41,7 @@ help:
 	@echo '  all (default)    - Build and test all'
 	@echo '  build            - Build all'
 	@echo '  install          - Build and install artifacts'
+	@echo '  docker           - Build a Docker image'
```
sergefdrv, line 45: Maybe also add `docker-clean` target to remove the docker image?

sergefdrv (commented):
Thank you, this PR looks to me much better now! I just left some minor comments.

README.md
ynamiki: I will add `sudo` and delete "if ..., and the current user can run docker command"; I am using `sudo` in other `docker` command examples, so I think it will has consistency in the style by doing so.

ynamiki:
@Naoya-Horiguchi @sergefdrv Thanks for comments. I have updated.

Changes from the previous revision ([diffs](https://github.com/hyperledger-labs/minbft/compare/d6194e40f2e83fefdf70e475a22f49a06d1d91b1..b357eddea89643758b4c078a2e249374fc78767f)):
- Fix some errors in English grammar in README.
- Add `docker-clean` target in Makefile to delete the image.
- Add `User` option to docker-compose.yml to run the process in the container as not `root` but a user who has the same UID as the current user in the host. By this, the owner of keys.yaml, which is created in the container in the directory shared with host, is now the current user in the host. So we don't need `sudo` for deleting the file (as described in the section Tear Down). Note that, `$UID` is not an environment variable but a shell variable of `bash`, we have to pass it to docker-compose by `UID=$UID docker-compose ...`. I have set the default value, 1000, which is the common first UID for non-root, non-system user in Ubuntu.

nhoriguchi (dismissed):
Looks good to me, thank you 👍

README.md
sergefdrv: When we doing `sudo docker` it's more or less clear what could happen to the host system. On the other hand, `sudo make` could execute arbitrary command on the host system, and a careful user would have to check the Makefile...

Makefile
```diff
@@ -24,7 +24,7 @@ prefix ?= sample
 bindir := $(prefix)/bin
 libdir := $(prefix)/lib
 
-.PHONY: all build install uninstall clean check test lint generate
+.PHONY: all build install uninstall docker clean check test lint generate
```
sergefdrv: Please add `docker-clean` here as well.

README.md
```diff
@@ -103,6 +109,116 @@ number of consenting nodes than PBFT.
...
+ - [Docker Compose](https://docs.docker.com/compose/) (tested with
+   version 1.17.1)
+
+If you are using Ubuntu 18.04, they can be installed as follows:
+
+```sh
+sudo apt-get install docker.io docker-compose
+```
+
+Note that TEE is not required to run the containers; they will run in
```
sergefdrv: s/TEE/SGX-enabled CPU/

sergefdrv (commented):
Sorry, I just left a couple of tiny comments.

ynamiki:
Thanks for suggestions! I have updated again ([diffs](https://github.com/hyperledger-labs/minbft/compare/b357edd..487679f8f941a81f123a7955a23fcff8e18b78f5)).

sergefdrv (approved):
:+1: Thanks!

nhoriguchi (approved):

PR #198, opened 2020-10-26
Logger rework

### LINKED ISSUE #42: Make some logging parameters configurable

### LINKED ISSUE #46: Enhance logging capabilities in client module
Logging capabilities of `client` package should be improved. It should become similar to what `core` does. `peer request` CLI should ultimately support the same logging flags as `peer run` (cf. #42).

luckychess:
This pull request improves logging capabilities of the MinBFT:
- Logger instance can be created internally as well as provided from the outside
- Options for both client and replica are reworked and used not for just logging parameters but for the client/replica parameters in general
- Logging options (log file and log level) are now available for the client

Resolves: #42.
Resolves: #46.


core/message-handling_test.go
```diff
@@ -653,7 +653,7 @@ func TestMakeGeneratedMessageConsumer(t *testing.T) {
 		return clientState
 	}
 
-	consume := makeGeneratedMessageConsumer(log, clientStates, logging.MustGetLogger(module))
+	consume := makeGeneratedMessageConsumer(log, clientStates, logger.NewReplicaLogger(clientID))
```
sergefdrv: It looks like `clientID` should be replaced with `0` here.

core/replica.go
```diff
@@ -46,6 +42,29 @@ type replica struct {
...
+type Option func(*options)
+
+func makeDefaultOptions(id uint32, opts ...Option) options {
+	l := commonLogger.NewReplicaLogger(id)
+	opt := options{logger: &l}
+	for _, o := range opts {
+		o(&opt)
+	}
+	return opt
+}
```
sergefdrv: It would be better to check `opt.logger` after applying `opts` and construct the default logger only if necessary.

core/request_test.go
```diff
@@ -390,7 +389,7 @@ func TestMakePrepareTimerStarter(t *testing.T) {
 	unicastLogs := map[uint32]messagelog.MessageLog{
 		uint32(0): messagelog.New(),
 	}
-	startTimer := makePrepareTimerStarter(uint32(0), provider, unicastLogs, logging.MustGetLogger(module))
+	startTimer := makePrepareTimerStarter(uint32(0), provider, unicastLogs, logger.NewReplicaLogger(clientID))
```
sergefdrv: It looks like `clientID` should be replaced with `0` here.

client/client.go
```diff
@@ -52,22 +46,46 @@ type Client interface {
...
+type Option func(*options)
+
+func makeDefaultOptions(id uint32, opts ...Option) options {
+	l := commonLogger.NewClientLogger(id)
+	opt := options{logger: &l}
+	for _, o := range opts {
+		o(&opt)
+	}
+	return opt
+}
```
sergefdrv: It would be better to check `opt.logger` after applying `opts` and construct the default logger only if necessary.

sergefdrv (commented):
Thank you for your contribution. I left some minor comments.

sergefdrv:
> Resolves: #42, #46.

It seems like this line will only resolve the first mentioned issue. I guess one needs to repeat "resolve" twice...

sergefdrv:
@luckychess As described in [contribution guideline](https://github.com/hyperledger-labs/minbft/blob/master/CONTRIBUTING.md#addressing-review-comments), for the sake of maintaining usable commit history, we prefer addressing review comments by directly changing the corresponding commits. Would you mind doing interactive rebase and force-pushing the branch?

common/logger/logger.go
```diff
@@ -0,0 +1,32 @@
...
+	Debugf(template string, args ...interface{})
+	Warning(args ...interface{})
+	Warningf(template string, args ...interface{})
+	Error(args ...interface{})
+	Errorf(template string, args ...interface{})
+	Info(args ...interface{})
+	Infof(template string, args ...interface{})
+	Fatal(args ...interface{})
+	Fatalf(template string, args ...interface{})
+}
```
sergefdrv: Why don't we order the methods by increasing severity level, i.e. debug, info, warning, error, fatal?

core/replica.go
```diff
@@ -46,6 +42,32 @@ type replica struct {
 	handleClientStream messageStreamHandler
 }
 
+type options struct {
+	logger *commonLogger.Logger
```
sergefdrv: `commonLogger.Logger` is an interface, which is a pointer type. What is a point in making `logger` field of a double-pointer type? Could we change it to `logger commonLogger.Logger`? Same applies for the parameter of `WithLogger`.

client/client.go
```diff
@@ -52,22 +46,49 @@ type Client interface {
 	Request(operation []byte) (resultChan <-chan []byte)
 }
 
+type options struct {
+	logger *commonLogger.Logger
```
sergefdrv: Same questions about the field type as in `core` package.

sergefdrv (changes requested):

sergefdrv:
@luckychess Thank you for addressing comments and updating the branch. I took another look and left some minor comments. I think we will merge this PR soon after your next update.

ynamiki (commented):
Thanks, I have the same comments as Sergey.

core/replica.go
luckychess: Sometimes my previous C++ experience reminds about it :)

sergefdrv (approved):

ynamiki (approved):
Thanks!

PR #204, opened 2021-05-14
Implement ViewChange message type

sergefdrv:
This pull request introduces and implements `ViewChange` message type. `ViewChange` message will be generated when handling `ReqViewChange` messages (#177).


messages/authen.go
```diff
@@ -70,6 +72,15 @@ func writeAuthenBytes(buf io.Writer, m Message) {
...
 		_ = binary.Write(buf, binary.BigEndian, m.NewView())
+	case ViewChange:
+		_ = binary.Write(buf, binary.BigEndian, m.NewView())
+		// There is no need to authenticate any other message
+		// content. Messages in the message log cannot be
+		// omitted/equivocated without it being detected,
+		// whereas any valid view-change certificate is
+		// equivalent to any other valid one. Moreover,
+		// messages in the message log and view-change
+		// certificate are authenticated on their own.
```
nhoriguchi, line 84: Maybe I miss something, but It seems to me unclear why we don't have to include message log into signed data. I think that each message has a signature, but the order of messages in the log is not signed, so could a malicious replica intentionally reorder the log and disturb the replay after view change?

nhoriguchi (commented):
I left one minor comment, but I can't find any critical issue on suggested changes, thank you.

messages/authen.go
sergefdrv: Messages in the message log are required to be sequential in the counter value assigned by USIG. So any reordering, omission, or duplication of messages in the log will be detected when validating integrity of the ViewChange message.

I will add this explanation to the comment.

sergefdrv:
Summary of [changes](https://github.com/hyperledger-labs/minbft/compare/5d0f23f182ae2859e0023fb1cac1030b7cff7016..b5ffd9a3459cd96224005ae2cb39ccaa071a1afe): adjusted comment in `messages/authen.go`.

nhoriguchi (approved):
Looks good to me.

sergefdrv:
> Looks good to me.

@nhoriguchi Thanks for reviewing it

PR #207, opened 2021-06-17
Support obtaining and resetting the message log

### LINKED ISSUE #177: Implement ReqViewChange message handling

sergefdrv:
This pull request extends the message log functionality required to generate `ViewChange` messages.

### DIFF
--- a/core/internal/messagelog/messagelog.go
+++ b/core/internal/messagelog/messagelog.go
@@ -43,7 +43,7 @@ type MessageLog interface {
 }
 
 type messageLog struct {
-	lock sync.RWMutex
+	sync.RWMutex
 
 	// Messages in order added
 	msgs []messages.Message
@@ -58,8 +58,8 @@ func New() MessageLog {
 }
 
 func (log *messageLog) Append(msg messages.Message) {
-	log.lock.Lock()
-	defer log.lock.Unlock()
+	log.Lock()
+	defer log.Unlock()
 
 	log.msgs = append(log.msgs, msg)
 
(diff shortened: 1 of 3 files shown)


core/internal/messagelog/messagelog.go
```diff
@@ -37,38 +37,44 @@ import (
...
+	Messages() []messages.Message
+	Reset(msgs []messages.Message)
 }
 
 type messageLog struct {
-	lock sync.RWMutex
+	sync.RWMutex
+
+	// Value incremented on each reset
+	epoch uint64
```
nhoriguchi: This field is used to detect `Reset()` calls during `supplyMessages()`. So naming this more directly like "reset" might be suitable? I think that the word "epoch" can be used to represent the similar concept as "view", so it might be confusing for someone familiar with other consensus protocols.

nhoriguchi (commented):
I left one nitpicking comment, but can't find any critical issues on the changes.

core/internal/messagelog/messagelog.go
sergefdrv: I decided to replace it with a channel.

sergefdrv:
Summary of [changes](https://github.com/hyperledger-labs/minbft/compare/538c8145756e905371e7b20f7b0448d8cf476d5c..c3ba51d40e1511ffac1144cc8df6b03e017cdb1f):
 - switched to using a channel for detecting log reset
 - improved testing of the new functionality

core/internal/messagelog/messagelog_test.go
```diff
@@ -80,6 +80,86 @@ func TestStream(t *testing.T) {
...
+					}
+				}, "Unexpected message from stream %d", streamID)
+				if !ok {
+					break
+				}
+			}
+		}(id)
+	}
+
+	log.Reset(msgs2)
```
nhoriguchi, line 159: It seems likely that the log is reset with msgs2 before the goroutines receive messages from msgs. So is it better to put a short delay before this reset?

sergefdrv: If we add a delay then the goroutines would likely receive all messages in `msgs` before the log is reset with `msgs2`.

nhoriguchi: OK, so maybe the delay is not a good enough idea, I have no better option so I'm fine with current code.

nhoriguchi (approved):
Looks good to me, thank you.

PR #212, opened 2021-11-10
Generalize Commit message type

sergefdrv:
This pull request generalizes Commit message type such that it can be used to reach agreement on the initial state of the new view and safely execute the same set of client requests that were potentially executed in the old view.


nhoriguchi (approved):
The changes look good to me, and the commit message clarifies the motivation and is convincing. Thank you.

PR #213, opened 2021-11-29
Perform full testing in make test target by default

sergefdrv:
This pull request changes the default behavior of make test target to perform full testing. The CI behavior is not changed: it still runs only short tests by setting `GOFLAGS="-short"` in the environment.


ynamiki (approved):
I cannot tell if some tests were skipped or not from the logs of GitHub Actions (if the short test is performed, a test in `sample/authentication/crypto_test.go` should be skipped). However, I confirmed that the environment variable `GOFLAGS` is effective in my local environment.

PR #214, opened 2021-11-29
NewView message type

sergefdrv:
This pull request defines and implements NewView message type. NewView messages will be generated when the new primary collects a quorum of ViewChange messages.

Related to #178.


messages/protobuf/testutils_test.go
```diff
@@ -72,5 +73,261 @@ func requireCertMsgEqual(t *testing.T, m1, m2 messages.CertifiedMessage) {
...
+				nvCert = append(nvCert, vcs[r])
+			}
+			out <- nvCert
+		}
+	}()
+
+	return out
+}
+
+// terminateMessageLogs replaces the gmessage logs in the supplied set,
```
nhoriguchi: nit: s/gmessage/message/ ?

nhoriguchi (commented):
I think that the change set implements `NewView` message straightforwardly (in the similar way to other message types, so looks friendly for reviewers). The test utility functions like `generateMessageLogs()` looks complicated at first, but I finally see that it generates many possible sets of message logs to prepare test data sets. This process is complicated in itself, so the generating function handles it efficiently. So, it's nice to me.

I admit that I still don't fully understand implementation details especially around `extendQuorum` part. You opened discussion #215 earlier, so maybe I will learn about it more.

Anyway, I left one minor comment about typo, but I almost agree with this PR, thank you.

sergefdrv:
> the change set implements `NewView` message straightforwardly

Yes, the implementation was optimized for simplicity. However, it is not efficient in terms of the serialized data size since it doesn't attempt to avoid duplication. We should consider optimizing it later.

sergefdrv:
> don't fully understand implementation details especially around `extendQuorum` part

`extendQuorum` starts with a provided set of replicas `q` and "recursively" extends it in different possible ways with `t` other replica IDs from the range _[0, n)_. If `t == 0` then it simply provides a copy of `q`. Otherwise it first provides replica sets extending `q` with `t` other replica IDs from the range _[0, n-1)_, then it provides replica sets extending `q` with `t-1` other replica IDs from the range _[0, n-1)_ and the replica ID `n-1`.

When `extendQuorum` is used in `generateNewViewCertificates`, the initial replica set includes the new primary `p`. The new primary's own ViewChange message must be present in the new-view certificate since it represents the new primary's message log prior to the NewView message. This allows any correct replica to proceed with processing an indirectly received NewView message without relying on the potentially faulty new primary to provide certified messages generated prior to the NewView message.

sergefdrv:
Summary of [changes](https://github.com/hyperledger-labs/minbft/compare/028c64ff8bfde8f1092fc94de9a356f17c56a377..6c7cc0f3975a0e3e20b3af90482caec7442c45aa): rebased and adjusted comments.

nhoriguchi:
> When `extendQuorum` is used in `generateNewViewCertificates`, the initial replica set includes the new primary `p`. The new primary's own ViewChange message must be present in the new-view certificate since it represents the new primary's message log prior to the NewView message. This allows any correct replica to proceed with processing an indirectly received NewView message without relying on the potentially faulty new primary to provide certified messages generated prior to the NewView message.

Thank you very much, this fully explains implicit protocol detail of view-change I missed.

nhoriguchi (approved):

PR #219, opened 2021-12-06
Simplified handling of certified messages

sergefdrv:
This pull request changes handling of messages certified with USIG. It requires replicas to send certified messages strictly sequentially in the assigned UI values. This applies both to replica's own messages and foreign messages embedded into replica's own messages. Moreover, only correct messages update the UI value tracked for each peer replica. These changes aim at simplifying reasoning about correctness of the implementation, especially when it comes to view change.

Related to #217, #178.


nhoriguchi (approved):
Looks good to me. So in current version, the message is considered as incorrect when `processViewMessage()` fails, which sounds clear to me.

PR #222, opened 2021-12-13
Specify commit certificate size explicitly

sergefdrv:
This pull request makes the commit certificate size explicit in the code.

Related to #215, #178.

### DIFF
--- a/core/req-view-change.go
+++ b/core/req-view-change.go
@@ -83,11 +83,11 @@ func makeReqViewChangeProcessor(collect reqViewChangeCollector, startViewChange 
 	}
 }
 
-func makeReqViewChangeCollector(f uint32) reqViewChangeCollector {
+func makeReqViewChangeCollector(commitCertSize, n uint32) reqViewChangeCollector {
 	var (
 		view      uint64
-		collected = make(messages.ViewChangeCert, 0, f+1)
-		replicas  = make(map[uint32]bool, f+1)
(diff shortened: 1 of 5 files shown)


core/req-view-change.go
```diff
@@ -83,11 +83,11 @@ func makeReqViewChangeProcessor(collect reqViewChangeCollector, startViewChange
 	}
 }
 
-func makeReqViewChangeCollector(f uint32) reqViewChangeCollector {
+func makeReqViewChangeCollector(commitCertSize, n uint32) reqViewChangeCollector {
 	var (
 		view      uint64
-		collected = make(messages.ViewChangeCert, 0, f+1)
-		replicas  = make(map[uint32]bool, f+1)
+		collected = make(messages.ViewChangeCert, 0, n-commitCertSize+1)
```
nhoriguchi, line 89: In my first look, I though this change is beyond the simple cleanup, but as you [mentioned](https://github.com/hyperledger-labs/minbft/discussions/215#discussioncomment-1822594) in #215 more than `f+1` ReqViewChange message can be required for generic case of `n > 2f + 1`. So this change is justified as such.

nhoriguchi (approved):
The patch looks good to me. Thank you.

core/req-view-change.go
sergefdrv: Yes, the old code assumed `n=2f+1`.

PR #227, opened 2022-01-05
Make message testing implementation-independent

sergefdrv:
This pull request introduces generic message testing utilities.

Related to #178.


nhoriguchi (approved):
Sorry for my slow response, the changeset looks good to me. Most of changes are renaming to change visibility of function but as stated in the commit log they can be called from outside package `testing` in the future, so I have no suspicion on it. Thank you.

PR #230, opened 2022-01-17
View change message handling

### LINKED ISSUE #178: Implement ViewChange message handling

sergefdrv:
This pull request implements handling of the ViewChange message type.

Resolves: #178

### DIFF
--- a/core/message-handling.go
+++ b/core/message-handling.go
@@ -181,7 +181,9 @@ func defaultMessageHandlers(id uint32, log messagelog.MessageLog, unicastLogs ma
 	validatePrepare := makePrepareValidator(n)
 	validateCommit := makeCommitValidator()
 	validateReqViewChange := makeReqViewChangeValidator(verifyMessageSignature)
-	validateCertifiedMessage := makeCertifiedMessageValidator(validatePrepare, validateCommit, verifyUI)
+	validateVCCert := makeViewChangeCertValidator(viewChangeCertSize)
+	validateViewChange := makeViewChangeValidator(validateMessageLog, validateVCCert)
+	validateCertifiedMessage := makeCertifiedMessageValidator(validatePrepare, validateCommit, validateViewChange, verifyUI)
 	validatePeerMessage := makePeerMessageValidator(n, validateCertifiedMessage, validateReqViewChange)
 	validateMessage := makeMessageValidator(validateRequest, validatePeerMessage)
 
@@ -483,7 +485,7 @@ func makePeerMessageValidator(n uint32, validateCertified certifiedMessageValida
 	}
 }
 
-func makeCertifiedMessageValidator(validatePrepare prepareValidator, validateCommit commitValidator, verifyUI uiVerifier) certifiedMessageValidator {
+func makeCertifiedMessageValidator(validatePrepare prepareValidator, validateCommit commitValidator, validateViewChange viewChangeValidator, verifyUI uiVerifier) certifiedMessageValidator {
 	return func(msg messages.CertifiedMessage) error {
 		if err := verifyUI(msg); err != nil {
 			return fmt.Errorf("invalid UI: %s", err)
@@ -494,6 +496,8 @@ func makeCertifiedMessageValidator(validatePrepare prepareValidator, validateCom
 			return validatePrepare(msg)
 		case messages.Commit:
 			return validateCommit(msg)
+		case messages.ViewChange:
+			return validateViewChange(msg)
 		default:
 			panic("Unknown message type")
 		}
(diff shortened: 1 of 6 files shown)


core/message-handling.go
```diff
@@ -177,7 +177,9 @@ func defaultMessageHandlers(id uint32, log messagelog.MessageLog, unicastLogs ma
 	validatePrepare := makePrepareValidator(n)
 	validateCommit := makeCommitValidator()
 	validateReqViewChange := makeReqViewChangeValidator(verifyMessageSignature)
-	validateCertifiedMessage := makeCertifiedMessageValidator(validatePrepare, validateCommit, verifyUI)
+	validateVCCert := makeViewChangeCertValidator(commitCertSize, n)
+	validateViewChange := makeViewChangeValidator(validateMessageLog, validateVCCert)
```
nhoriguchi, line 185: `validateMessageLog()` seems to be statically defined in `core/view-change.go`, so we don't have to explicitly pass the function via argument?

core/view-change.go
```diff
@@ -0,0 +1,130 @@
...
+			return fmt.Errorf("invalid message log: %s", err)
+		}
+
+		return nil
+	}
+}
+
+func makeViewChangeCertValidator(commitCertSize, n uint32) viewChangeCertValidator {
+	return func(newView uint64, cert messages.ViewChangeCert) error {
+		if len(cert) <= int(n-commitCertSize) {
```
nhoriguchi: I guess that theoretical minimum requirement is `commitCertSize` and you intentionally use the stronger requirement of `n - commitCertSize + 1` based on discussion #215. If so,  is it worth adding comment (or leaving pointer to the #215 is also fine) about the choice?

core/view-change.go
```diff
@@ -128,3 +153,71 @@ func validateMessageLog(replicaID uint32, view uint64, log messages.MessageLog,
...
+			view = newView
+		}
+
+		if !isPrimary(view, id, n) {
+			return false, nil
+		}
+
+		if vc.ReplicaID() == id {
+			collected[0] = vc
+		} else if len(collected) <= int(n-commitCertSize) {
```
nhoriguchi: This kind of check is called multiple times, so how about factoring out to a separate function with a descriptive name?

core/message-handling.go
sergefdrv: This dependency injection is useful for unit-testing

core/view-change.go
sergefdrv: Any view-change certificate must intersect with any commit certificate, otherwise delayed/slow replicas may keep triggering view change, thus the threshold of `n-commitCertSize+1`.

core/view-change.go
```diff
@@ -128,3 +153,71 @@ func validateMessageLog(replicaID uint32, view uint64, log messages.MessageLog,
...
+		}
+
+		return
+	}
+}
+
+func makeViewChangeCollector(id, commitCertSize, n uint32) viewChangeCollector {
+	var (
+		view      uint64
+		collected = make(messages.NewViewCert, 1, n-commitCertSize+1)
```
sergefdrv: Oh, the capacity should be `n-commitCertSize`

core/view-change.go
sergefdrv: Maybe just define `threshold := n-commitCertSize`?

sergefdrv: I think the condition here should be `len(collected) < int(n-commitCertSize)` resp. `len(collected) < int(threshold)`, need to check the unit-test...

sergefdrv: Ah, no, `commitCertSize == f+1`, so the condition is correct. Then we would define `threshold := n-commitCertSize+1`.

core/view-change.go
sergefdrv: The capacity is correct, see [here](https://github.com/hyperledger-labs/minbft/pull/230/files/45aa5e4aa2cf5a145e64a9af6c67b366cab016c1..db5ca76d3b2750be3a942fe9bd1285d21b227898#r789942736).

core/message-handling.go
nhoriguchi: OK, thanks.

core/view-change.go
nhoriguchi: Yes, so if this insight is from your own (not from the original paper), l thought that leaving some comment about it around the code is helpful for future developers.

core/view-change.go
nhoriguchi: Thank you for double checking, `threshold := n-commitCertSize` looks fine to me.

core/view-change.go
sergefdrv: Indeed, I will add some comment

core/view-change.go
sergefdrv: Perhaps `viewChangeCertSize` would be a better name. What do you think about passing it to `makeViewChangeCertValidator`, `makeViewChangeCollector`, and `makeReqViewChangeCollector` as a parameter?

nhoriguchi: Sounds nice 👍

sergefdrv:
Summary of [changes](https://github.com/hyperledger-labs/minbft/compare/db5ca76d3b2750be3a942fe9bd1285d21b227898..4103c85aa7e8cbeeb14be04142044ef177e47812): added explicit definition for view-change certificate size.

nhoriguchi (approved):
Thank you very much.

PR #233, opened 2022-02-04
NewView message handling prep 2: un-preparing requests

sergefdrv:
This pull request adjusts the client state upon starting view change.

Related to #179

### DIFF
--- a/core/message-handling.go
+++ b/core/message-handling.go
@@ -158,6 +158,7 @@ func defaultMessageHandlers(id uint32, log messagelog.MessageLog, unicastLogs ma
 	captureSeq := makeRequestSeqCapturer(clientStates)
 	prepareSeq := makeRequestSeqPreparer(clientStates)
 	retireSeq := makeRequestSeqRetirer(clientStates)
+	unprepareSeq := makeRequestSeqUnpreparer(clientStates)
 	pendingReq := requestlist.New()
 
 	consumeGeneratedMessage := makeGeneratedMessageConsumer(log, clientStates, logger)
@@ -200,7 +201,7 @@ func defaultMessageHandlers(id uint32, log messagelog.MessageLog, unicastLogs ma
 	processCertifiedMessage := makeCertifiedMessageProcessor(n, processViewMessage, processViewChange)
 
 	collectReqViewChange := makeReqViewChangeCollector(viewChangeCertSize)
-	startViewChange := makeViewChangeStarter(id, viewState, log, handleGeneratedMessage)
+	startViewChange := makeViewChangeStarter(id, viewState, log, unprepareSeq, handleGeneratedMessage)
 	processReqViewChange := makeReqViewChangeProcessor(collectReqViewChange, startViewChange)
 
 	processPeerMessage := makePeerMessageProcessor(n, processCertifiedMessage, processReqViewChange)
--- a/core/req-view-change.go
+++ b/core/req-view-change.go
@@ -113,7 +113,7 @@ func makeReqViewChangeCollector(viewChangeCertSize uint32) reqViewChangeCollecto
 	}
 }
 
-func makeViewChangeStarter(id uint32, viewState viewstate.State, log messagelog.MessageLog, handleGeneratedMessage generatedMessageHandler) viewChangeStarter {
+func makeViewChangeStarter(id uint32, viewState viewstate.State, log messagelog.MessageLog, unprepareReqSeq requestSeqUnpreparer, handleGeneratedMessage generatedMessageHandler) viewChangeStarter {
 	return func(newView uint64, vcCert messages.ViewChangeCert) (ok bool, err error) {
 		ok, release := viewState.AdvanceExpectedView(newView)
 		if !ok {
@@ -130,6 +130,7 @@ func makeViewChangeStarter(id uint32, viewState viewstate.State, log messagelog.
 		log.Reset(nil)
 
 		// TODO: start view-change timer
+		unprepareReqSeq()
 
 		handleGeneratedMessage(messageImpl.NewViewChange(id, newView, msgs, vcCert))
 
--- a/core/req-view-change_test.go
+++ b/core/req-view-change_test.go
@@ -155,7 +155,10 @@ func TestMakeViewChangeStarter(t *testing.T) {
 	handleGenerated := func(msg messages.ReplicaMessage) {
 		mock.MethodCalled("generatedMessageHandler", msg)
 	}
-	start := makeViewChangeStarter(id, viewState, log, handleGenerated)
+	unprepareSeq := func() {
+		mock.MethodCalled("requestSeqUnpreparer")
+	}
+	start := makeViewChangeStarter(id, viewState, log, unprepareSeq, handleGenerated)
 
 	newView := rand.Uint64()
 	cert := messages.ViewChangeCert{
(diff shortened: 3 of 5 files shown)


nhoriguchi (dismissed):

sergefdrv:
After a second thought, I'm not anymore sure that `viewChangeStarter` is the right place to un-prepare requests because some requests may [get prepared again when processing delayed Prepare messages](https://github.com/sergefdrv/minbft/blob/view-change-unprepare/core/prepare.go#L65).

sergefdrv:
> some requests may [get prepared again when processing delayed Prepare messages](https://github.com/sergefdrv/minbft/blob/view-change-unprepare/core/prepare.go#L65)

Perhaps, `peerMessageApplier` should only be called in [`viewMessageProcessor`](https://github.com/sergefdrv/minbft/blob/view-change-unprepare/core/message-handling.go#L590-L636) if the view is active (`currentView == expectedView`).

sergefdrv:
> Perhaps, `peerMessageApplier` should only be called in [`viewMessageProcessor`](https://github.com/sergefdrv/minbft/blob/view-change-unprepare/core/message-handling.go#L590-L636) if the view is active (`currentView == expectedView`).

But then `viewMessageProcessor` may not reject some messages which other correct replicas will find incorrect; this can compromise liveness.

nhoriguchi:
> After a second thought, I'm not anymore sure that `viewChangeStarter` is the right place to un-prepare requests because some requests may [get prepared again when processing delayed Prepare messages](https://github.com/sergefdrv/minbft/blob/view-change-unprepare/core/prepare.go#L65).

I read the related code and felt that `viewChangeStarter` calls `viewState.AdvanceExpectedView()` which increments `expectedView` with holding write-mode lock. This write lock seems to be held until `viewChangeStarter` exits (which means whole view change process is done, I guess). On the other hand, `viewMessageProcessor` refers to the view status with `viewState.HoldView()` with read-mode lock, so it should wait here until the view change process is finished.  So, I think that the delayed Prepare messages can be expected to be rejected by `messageView < currentView` check (without jumping into `peerMessageApplier`)?

sergefdrv:
> ...`viewChangeStarter` exits (which means whole view change process is done, I guess)

Not exactly, `viewChangeStarter` is called when processing a ReqViewChange message and have to exit for handling the next message from the same message stream.

sergefdrv:
> > Perhaps, `peerMessageApplier` should only be called in [`viewMessageProcessor`](https://github.com/sergefdrv/minbft/blob/view-change-unprepare/core/message-handling.go#L590-L636) if the view is active (`currentView == expectedView`).
> 
> But then `viewMessageProcessor` may not reject some messages which other correct replicas will find incorrect; this can compromise liveness.

An easy solution was not to detect duplicate Prepare messages (preparing the same Request) at all, i.e. completely remove `requestSeqPreparer`. But I wanted to preserve this check here.

sergefdrv:
> > Perhaps, `peerMessageApplier` should only be called in [`viewMessageProcessor`](https://github.com/sergefdrv/minbft/blob/view-change-unprepare/core/message-handling.go#L590-L636) if the view is active (`currentView == expectedView`).
> 
> But then `viewMessageProcessor` may not reject some messages which other correct replicas will find incorrect; this can compromise liveness.

I think there is a solution for this problem. As mentioned [here](https://github.com/hyperledger-labs/minbft/discussions/217#discussioncomment-1757461), it boils down to maintaining the following invariant: If a correct replica accepts a message then any other correct replica will accept the message. 

Luckily, there is another useful invariant: Correct replicas starting view change must ensure that all other correct replicas will do the same. This is achieved by justifying the ViewChange message with a view-change certificate (quorum of ReqViewChange messages). Therefore, if a correct replica does not reject some message because it has started a view change then any other correct replica will do the same when received the message relayed (embedded) by that replica, since it must also have started the view change before processing the relayed message.

Thus I consider skipping `requestSeqPreparer` in `prepareApplier` when the view is inactive (`active == false`), i.e. only check for duplicate Prepare messages when the current view is active. Correct replicas will still reject duplicate Prepare messages in an active view.

sergefdrv:
Summary of [changes](https://github.com/hyperledger-labs/minbft/compare/d7e947a2c01d95619707995d17fe2dd666095e1f..7b329218cda537faf754473be9c6cd72a6fffb81):
 - added new commit: "[core: Prepare request identifiers only in active view](https://github.com/hyperledger-labs/minbft/pull/233/commits/97da2c912b0a8995930c2cf4d32574c5d76ba54f)"

sergefdrv:
> > > Perhaps, `peerMessageApplier` should only be called in [`viewMessageProcessor`](https://github.com/sergefdrv/minbft/blob/view-change-unprepare/core/message-handling.go#L590-L636) if the view is active (`currentView == expectedView`).
> > 
> > 
> > But then `viewMessageProcessor` may not reject some messages which other correct replicas will find incorrect; this can compromise liveness.
> 
> I think there is a solution for this problem. As mentioned [here](https://github.com/hyperledger-labs/minbft/discussions/217#discussioncomment-1757461), it boils down to maintaining the following invariant: If a correct replica accepts a message then any other correct replica will accept the message.
> 
> Luckily, there is another useful invariant: Correct replicas starting view change must ensure that all other correct replicas will do the same. This is achieved by justifying the ViewChange message with a view-change certificate (quorum of ReqViewChange messages). Therefore, if a correct replica does not reject some message because it has started a view change then any other correct replica will do the same when received the message relayed (embedded) by that replica, since it must also have started the view change before processing the relayed message.
> 
> Thus I consider skipping `requestSeqPreparer` in `prepareApplier` when the view is inactive (`active == false`), i.e. only check for duplicate Prepare messages when the current view is active. Correct replicas will still reject duplicate Prepare messages in an active view.

Hmm, it looks like the same argument applies to `peerMessageApplier`, i.e. it can be skipped if the current view is not active.

sergefdrv:
> > > > Perhaps, `peerMessageApplier` should only be called in [`viewMessageProcessor`](https://github.com/sergefdrv/minbft/blob/view-change-unprepare/core/message-handling.go#L590-L636) if the view is active (`currentView == expectedView`).
> > > 
> > > 
> > > But then `viewMessageProcessor` may not reject some messages which other correct replicas will find incorrect; this can compromise liveness.
> > 
> > 
> > I think there is a solution for this problem. As mentioned [here](https://github.com/hyperledger-labs/minbft/discussions/217#discussioncomment-1757461), it boils down to maintaining the following invariant: If a correct replica accepts a message then any other correct replica will accept the message.
> > Luckily, there is another useful invariant: Correct replicas starting view change must ensure that all other correct replicas will do the same. This is achieved by justifying the ViewChange message with a view-change certificate (quorum of ReqViewChange messages). Therefore, if a correct replica does not reject some message because it has started a view change then any other correct replica will do the same when received the message relayed (embedded) by that replica, since it must also have started the view change before processing the relayed message.
> > Thus I consider skipping `requestSeqPreparer` in `prepareApplier` when the view is inactive (`active == false`), i.e. only check for duplicate Prepare messages when the current view is active. Correct replicas will still reject duplicate Prepare messages in an active view.
> 
> Hmm, it looks like the same argument applies to `peerMessageApplier`, i.e. it can be skipped if the current view is not active.

I implemented this approach in #235.

nhoriguchi (dismissed):

sergefdrv:
Summary of [changes](https://github.com/hyperledger-labs/minbft/compare/7b329218cda537faf754473be9c6cd72a6fffb81..dae9cc7329beeb0b4034e96fc74d59cdec9c2a99): rebased on top of a9ffc14 and dropped 97da2c9.

nhoriguchi (approved):

PR #234, opened 2022-02-15
Allow capturing USIG identity from any UI

sergefdrv:
This pull request adjusts the way USIG identity is captured.

Related to #179.

### DIFF
--- a/sample/authentication/crypto.go
+++ b/sample/authentication/crypto.go
@@ -198,17 +198,16 @@ func (au *SGXUSIGAuthenticationScheme) VerifyAuthenticationTag(m []byte, sig []b
 	au.lock.Lock()
 	defer au.lock.Unlock()
 
-	// Capture the epoch value received with the first valid UI to
-	// dynamically determine the full USIG identity. Note that
-	// this relies on the assumption that all peer replicas are
-	// initially correct, each uses a unique USIG key pairs per
-	// consensus protocol instance and generates its first UI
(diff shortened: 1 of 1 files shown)


sample/authentication/crypto.go
```diff
@@ -217,7 +217,7 @@ func (au *SGXUSIGAuthenticationScheme) VerifyAuthenticationTag(m []byte, sig []b
 	// UIs. This, for example, can be achieved using some
 	// bootstrapping procedure.
 	epoch, ok := au.epoch[fingerprint]
-	if !ok && ui.Counter == uint64(1) {
+	if !ok {
```
nhoriguchi, line 219: The above inline comment seems to mention the assumption about `ui.Counter == uint64(1)`. So does it need some update?

sergefdrv: Yes, it seems so. I think we should distinguish here "first received UI" from "first generated UI". Is it then as simple as changing "generates its first UI" to "generates its UIs"?

sergefdrv:
Summary of [changes](https://github.com/hyperledger-labs/minbft/compare/9640f09cc7b558f3d580295aacf44cff0e37557a..4208405688182614adc0f699eca10b57c7da92e8): updated comment.

sample/authentication/crypto.go
nhoriguchi: Thank you for the update.

nhoriguchi (approved):

PR #235, opened 2022-02-16
Apply only messages referring to the current active view

sergefdrv:
This pull request changes the way view messages are applied such that only messages referring to the current active view get applied.

Related to #179, #233.


nhoriguchi (approved):
Looks good to me, thank you :+1:

PR #237, opened 2022-02-18
Remove only matching requests from the request list

sergefdrv:
This pull request adjust the implementation of the request list internal component so that to avoid inadvertently removing more recent requests from the list.

Related to #179.


nhoriguchi (approved):
Looks good to me, thank you.

PR #238, opened 2022-02-18
Ignore Reply messages generated for old requests

sergefdrv:
This pull request removes excessive error checking when handling generated Reply messages.

Relates to #179.


nhoriguchi (approved):
Looks good to me.

PR #240, opened 2022-03-14
Add missing unit-test for makeNewViewAcceptor

sergefdrv:
This pull request adds a unit-test for `makeNewViewAcceptor` function that was missed in #239.

Related to #239, #179.

PR #241, opened 2022-03-14
View change timer

### LINKED ISSUE #180: Implement view change timer
The view change timer ensures a subsequent view change triggered if the new primary candidate is faulty.

sergefdrv:
This pull request implements view-change timer and thus delivers a complete implementation of the view change mechanism.

Resolves: #180.

### DIFF
--- a/core/internal/backofftimer/timer.go
+++ b/core/internal/backofftimer/timer.go
@@ -1,0 +1,89 @@
+// Copyright (c) 2022 NEC Laboratories Europe GmbH.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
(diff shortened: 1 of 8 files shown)


sergefdrv:
NB. This PR is based on top of #240.

core/internal/backofftimer/timer.go
```diff
@@ -0,0 +1,89 @@
...
+	return &timer{
+		options:        opts,
+		initTimeout:    d,
+		currentTimeout: d,
+	}
+}
+
+func (t *timer) Start(f func()) {
+	t.timer = t.timerProvider.AfterFunc(t.currentTimeout, f)
+	t.currentTimeout *= 2
```
nhoriguchi, line 79: Just from curiosity, what's the intention of this `*= 2`? This seems not to be visible to user of the timer, so no harm for now. Maybe it's useful for future changes?

sergefdrv: This line doubles the back-off timer's timeout on each restart. The timeout is reset back to the initial value when the timer is explicitly stopped. When the back-off timer is used as the view-change timer, it is in accordance with the paper:

> If additional view changes are needed, the timer is multiplied by two each time, increasing exponentially until a new primary responds. The objective is to avoid timer expirations forever due to long communication delays.

sergefdrv: We might consider adding an upper bound on the view-change timeout value as a protocol parameter in future.

nhoriguchi: It's clear now, thank you very much.

nhoriguchi (approved):

Issue #176, opened 2020-05-05
Implement view change


sergefdrv:
This is an umbrella issue that covers work related to implement view change mechanism. View change allows the system to survive primary replica (leader node) failure. The main work items:

- [x] #177
- [x] #178
- [x] #179 
- [x] #180
- [x] #243

PR #242, opened 2022-03-24
Fix handling messages embedded into own generated messages

sergefdrv:
This pull request fixes a bug in handling of own generated messages. A replica could skip handling some of the messages it has generated shortly before resetting its message log and replacing it with a newly generated ViewChange or NewView message.

Relates to #176.


nhoriguchi (approved):
Looks good to me.

Issue #216, opened 2021-12-01
Optimize message implementation for serialized data size


sergefdrv:
The message implementation was optimized for simplicity. However, it is not efficient in terms of the serialized data size since it doesn't attempt to avoid duplication.

PR #246, opened 2022-03-28
Semi-automated view-change test

### LINKED ISSUE #243: Add semi-automated view-change test
Implement a script that periodically suspends and resumes replica processes. This will help to better test view-change functionality.

sergefdrv:
This pull request resolves #243.

### DIFF
--- a/testing/suspend-resume.sh
+++ b/testing/suspend-resume.sh
@@ -1,0 +1,156 @@
+#!/bin/bash
+#
+# Copyright (c) 2022 NEC Laboratories Europe GmbH.
+#
+# Authors: Sergey Fedorov <sergey.fedorov@neclab.eu>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


sergefdrv:
The test sometimes fails. One issue is that clients continue submitting requests when replicas are stopped, so some replicas get stopped before they have chance to deliver all the requests being processed. I have a fix for that. But the test also happen to fail from time to time because it takes too long to handle messages communicated during view change. Remember that checkpoints are not implemented yet to limit the size of the protocol history (#218). Implementation of ViewChange and NewView messages is also simple, but inefficient (#216).

sergefdrv:
Summary of [changes](https://github.com/hyperledger-labs/minbft/compare/40fe97abf4137f720369a82a8d067a87cbb43288..1a2000ba3eac4e9649164f5bc6b1542d5da3aa71):
 - reduced default number of rounds
 - stopping clients before replicas

sergefdrv:
The test is much more stable with lower number of rounds. It may still fail because view-change messages sometimes get too big and exceed gRPC limit. We may want to increase that limit in our sample application.

nhoriguchi (approved):
Looks good to me.  And I ran the script and confirmed to pass. This is also a good demonstration of view-change process.

PR #259, opened 2023-11-06
update .golangci.yml

nhoriguchi:
I do want to merge #258 to archive this project, but I can't because of failure in GitHub Action (some linters spit warnings like "unused-parameter" and "redefines-builtin-id").  No one cares about CI failures of the project to be archived, so I forcibly disable related linters to ignore them.


ynamiki (approved):
Thanks for the updates.

